PR #2, opened 2016-11-25
Basic PerfMap implementation

schu:
Signed-off-by: Michael Schubert <michael@kinvolk.io>


alban:
It looks good to me but I would like if @drzaeus77 could check.

alban:
Merging for now. If there are more feedback, they can be addressed in future PRs.

PR #5, opened 2016-12-08
Remove `bpf` prefix from files and types + some cleanup and godoc

schu:


schu:
@drzaeus77 @alban ok to merge?

PR #7, opened 2017-01-06
elf: add map update and lookup functions

iaguis:


elf/elf.go
```diff
@@ -813,6 +834,58 @@ func perfEventPoll(fds []C.int) error {
...
+	)
+	ret, _, err := syscall.Syscall(
+		C.__NR_bpf,
+		C.BPF_MAP_UPDATE_ELEM,
+		uintptr(unsafe.Pointer(&uba)),
+		unsafe.Sizeof(uba),
+	)
+
+	if ret != 0 || err != 0 {
+		return fmt.Errorf("Unable to update element: %s", err)
```
alban: s/Unable/unable

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

ditto in the other function

alban:
The file `elf.go` starts to get big. I think the new feature should go in another file `table.go` like it is done in the bcc/ directory.

Ideally the next step will be to move some helper functions in a different files (e.g. `EnableKprobe()` which is not related to the elf loader)

elf/elf.go
```diff
@@ -696,9 +717,9 @@ func (b *Module) Load() error {
 			}
 
 			// assign perf fd tp map
-			ret := C.bpf_update_element(C.int(b.maps[name].m.fd), unsafe.Pointer(&cpu), unsafe.Pointer(&pmuFD), C.BPF_ANY)
-			if ret != 0 {
-				return fmt.Errorf("cannot assign perf fd to map: %d (cpu %d)", syscall.Errno(ret), cpu)
+			_, err = C.bpf_update_element(C.int(b.maps[name].m.fd), unsafe.Pointer(&cpu), unsafe.Pointer(&pmuFD), C.BPF_ANY)
```
alban: I see the cgo magic makes the function `C.bpf_update_element()` both the return value of `bpf()` system call and errno.

However, in case of success, `errno` might not be reset to zero so it might keep a non-zero value from a previous unrelated function.

man 3 errno:
> Its value is significant only when the return value of the call indicated an error

"man 2 bpf" does not promise to set errno to zero in case of success.

So I think the error check below is not correct.

iaguis: we can keep`if ret != 0`

iaguis:
Addressed the comment and did some initial refactoring.

PR #8, opened 2017-01-10
elf: special case current version for Ubuntu/Debian

schu:
Signed-off-by: Michael Schubert <michael@kinvolk.io>

### DIFF
--- a/elf/kernel_version_test.go
+++ b/elf/kernel_version_test.go
@@ -1,0 +1,52 @@
+// Copyright 2017 Kinvolk
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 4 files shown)


elf/kernel_version_test.go
```diff
@@ -0,0 +1,52 @@
...
+	{true, "4.1.2rkt-v1", 262402},
+	{true, "4.1.2-3 foo", 262402},
+	{false, "foo 4.1.2-3", -1},
+	{true, "4.1.2", 262402},
+	{false, ".4.1.2", -1},
+	{false, "4.1.", -1},
+	{false, "4.1", -1},
+}
+
+func TestKernelVersionFromReleaseString(t *testing.T) {
```
alban: could we have TravisCI or something running this test? This test does not need root so we could pick any CI.

schu: Happy to do this separately: #10

alban:
LGTM

PR #17, opened 2017-01-23
perf event reordering: improve API

alban:
Before this patch, the perf events reordering code was assuming that
there was a timestamp in the first field of the perf event struct. That
convention was not explicit.

This patch removes this assumption: by default, perf events are not
ordered unless SetGetTimestamp() is called to register a callback
getting the timestamp.

-----

This patch was tested with the branch https://github.com/kinvolk/scope/tree/alban/reordering-api

@iaguis @schu @alepuccetti PTAL

### DIFF
--- a/elf/perf.go
+++ b/elf/perf.go
@@ -108,6 +108,7 @@ type PerfMap struct {
 	program      *Module
 	receiverChan chan []byte
 	pollStop     chan bool
+	timestamp    func(*[]byte) uint64
 }
 
 // Matching 'struct perf_event_sample in kernel sources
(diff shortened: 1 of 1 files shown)


elf/perf.go
```diff
@@ -131,8 +132,20 @@ func InitPerfMap(b *Module, mapName string, receiverChan chan []byte) (*PerfMap,
...
 
+// SetGetTimestamp registers a getTimestamp callback that will be used to
+// reorder the perf events chronologically.
+//
+// If not set, the order of events sent through receiverChan is not guaranteed.
+//
+// Typically, the ebpf program will use bpf_ktime_get_ns() to get a timestamp
+// and store it in the perf event. The perf event struct is opaque to this
+// package, hence the need for a callback.
+func (pm *PerfMap) SetGetTimestamp(getTimestamp func(*[]byte) uint64) {
```
iaguis: Maybe call it `SetTimestampFunc`

elf/perf.go
```diff
@@ -108,6 +108,7 @@ type PerfMap struct {
 	program      *Module
 	receiverChan chan []byte
 	pollStop     chan bool
+	getTimestamp func(*[]byte) uint64
```
iaguis: Maybe call this `timestamp()`

alban:
@iaguis renamed. I also updated the Scope branch I used for testing.

iaguis:
I was wondering if we should add `error` to the return of `timestamp()` but I guess it's not worth it?

alban:
> I was wondering if we should add `error` to the return of `timestamp()` but I guess it's not worth it?

I am not sure about this. To me, it would look like a programmatic error and the user of the gobpf lib could either panic or print an error message. But I don't think there is any action that gobpf should do whenever such an error happens.

iaguis:
> I am not sure about this. To me, it would look like a programmatic error and the user of the gobpf lib could either panic or print an error message. But I don't think there is any action that gobpf should do whenever such an error happens.

Yes, let's leave it like it is.

lgtm

PR #18, opened 2017-01-24
Add dummy SetTimestampFunc for unsupported systems

iaguis:


elf/perf_unsupported.go
```diff
@@ -10,6 +10,8 @@ func InitPerfMap(b *Module, mapName string, receiverChan chan []byte) (*PerfMap,
 	return nil, fmt.Errorf("not supported")
 }
 
+func (pm *PerfMap) SetTimestampFunc(timestamp func(*[]byte) uint64) {}
```
alban: does it compile?

It should have a "return" statement because the function does not return void.

alban: err, sorry, I cannot read. SetTimestampFunc does not return anything.

iaguis: confusing function variables :)

PR #23, opened 2017-02-15
elf: don't assume all maps are perf maps

iaguis:
cc @alban

### DIFF
--- a/elf/elf.go
+++ b/elf/elf.go
@@ -478,11 +478,19 @@ func (b *Module) Load() error {
 		}
 	}
 
-	for name, _ := range b.maps {
+	return b.initializePerfMaps()
+}
+
+func (b *Module) initializePerfMaps() error {
+	for name, m := range b.maps {
 		var cpu C.int = 0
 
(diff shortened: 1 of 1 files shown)


alban:
This issue was discovered via https://github.com/weaveworks/scope/pull/2135#issuecomment-279901474
Symptoms:
> error initializing tcptracer_status map: unable to update element: argument list too long

alban:
CI fails.

@schu any idea what's happening? I don't see obvious error.

schu:
>  any idea what's happening? I don't see obvious error.

No. Green now, after triggering a new build.

elf/elf.go
```diff
@@ -478,9 +478,13 @@ func (b *Module) Load() error {
 		}
 	}
 
-	for name, _ := range b.maps {
+	for name, m := range b.maps {
```
alban: could we move this block in a separate function? So the name of the function would make it clear this is about initializing the perf maps, and not just generic elf parsing

iaguis: Done

PR #24, opened 2017-02-16
README: add header and semaphore ci badge

schu:


alepuccetti:
LGTM

alban:
It is annoying that the SemaphoreCI URL refers to my personal account rather than the organization. But this is better than nothing, while we figure out a better way.

Could you add documentation (probably in semaphore.sh):
- how SemaphoreCI was configured (which platform etc.)
- that semaphore.sh was taken from https://github.com/kinvolk/stage1-builder
So gobpf forks could easily setup their own CI if they want to.
This is explained in the other repositories but not in this one afaik.

But this can be done in a future PR so LGTM.

PR #27, opened 2017-02-23
Update gobpf/bcc to work with bcc 0.3.0

schu:


schu:
`bpf_attach_kprobe` signature was changed with iovisor/bcc@fdd2389c6a93307742d520a1e77d7e052bafcb27 ff

WIP as there's no release yet with the new signature.

mvbpolito:
Hi @schu, 

What do you think about doing a gobpf release before this commit and them merging this into master?

I ask this because we are in a position where we have to use a new bcc release (in order to use LPM-trie maps) and also gopbf.

Finally, just to ask, is there any particular reason to keep this library in its own repo instead of being in github.com/iovisor/bcc/tree/master/src?

Thanks,
Mauricio

schu:
Hey @mvbpolito 

I just updated the outdated [gobpf-ci Docker image](https://github.com/schu/docker-gobpf-ci/blob/892037e398a7f313980cae59799d4c2a55c8962a/Dockerfile) (switched from nightly to main builds) and changed gobpf/bcc to work with bcc 0.3.0.

> What do you think about doing a gobpf release before this commit and them merging this into master?

I meant to do regular tagged releases, but don't think we have to or should wait with merging this pull request.

>  is there any particular reason to keep this library in its own repo instead of being in github.com/iovisor/bcc/tree/master/src?

One reason is that gobpf is not pure bcc bindings with gobpf/elf.

schu:
@alban please take a look.

alban:
LGTM

PR #34, opened 2017-03-22
bcc: cast reader explicitly

iaguis:
Newer versions of go (1.8+) are more strict about this, so the
compilation failed:

```
github.com/iovisor/gobpf/bcc
# github.com/iovisor/gobpf/bcc
bcc/perf.go:131: cannot use reader (type unsafe.Pointer) as type *C.struct_perf_reader in argument to func literal
```

Cast it explicitly.


schu:
Thanks!

PR #36, opened 2017-03-22
Add support for uprobes

louism517:
Hi,

This PR adds supports for tracing uprobes.

It includes some new functions used to query the symbol cache for the location (file and offset) of function within userland code; and to retrieve a list of matching symbols (functions) from a given module (library or binary) to allow attaching to all symbols that match a certain pattern. (Both these ideas borrowed heavily from the Python implementation).

It also adds functions to the Module type to allow adding of uprobes, uretprobes, and all those that match a pattern.

Finally an example, strlen_count.go, of how to use these new functions.

I do have a query/problem that I could do with some advice on, regarding the example:

I have used a Table without a PerfMap (in contrast to the other example which is given). The tracing happens until the user hits Ctrl-C,and then we iterate over the table entries with the Table.Iter() function.

The problem I have is that results are returned as an Event{Key: string, Value: string} where Key and Value appear to be string representations of C *char strings, like:

"[ 0x66 0x76 0x33 0x0 ]" - i.e. this is a string, _not_ a series of chars. The only (ugly) way I could figure out to extract the actual string was to remove the 0x, concatenate each value and then use hex.DecodeString.

Is there a better way?

Many Thanks
Louis

### DIFF
--- a/bcc/module.go
+++ b/bcc/module.go
@@ -38,6 +38,7 @@ type Module struct {
 	p       unsafe.Pointer
 	funcs   map[string]int
 	kprobes map[string]unsafe.Pointer
+	uprobes map[string]unsafe.Pointer
 }
 
 type compileRequest struct {
@@ -89,6 +90,7 @@ func newModule(code string, cflags []string) *Module {
 		p:       c,
 		funcs:   make(map[string]int),
 		kprobes: make(map[string]unsafe.Pointer),
+		uprobes: make(map[string]unsafe.Pointer),
 	}
 }
 
@@ -118,6 +120,12 @@ func (bpf *Module) Close() {
 		C.bpf_detach_kprobe(evNameCS)
 		C.free(unsafe.Pointer(evNameCS))
 	}
+	for k, v := range bpf.uprobes {
+		C.perf_reader_free(v)
+		evNameCS := C.CString(k)
+		C.bpf_detach_uprobe(evNameCS)
+		C.free(unsafe.Pointer(evNameCS))
+	}
 	for _, fd := range bpf.funcs {
 		syscall.Close(fd)
 	}
@@ -130,6 +138,11 @@ func (bpf *Module) LoadNet(name string) (int, error) {
 
 // LoadKprobe loads a program of type BPF_PROG_TYPE_KPROBE.
 func (bpf *Module) LoadKprobe(name string) (int, error) {
+	return bpf.Load(name, C.BPF_PROG_TYPE_KPROBE)
+}
+
+// LoadUprobe loads a program of type BPF_PROG_TYPE_KPROBE.
+func (bpf *Module) LoadUprobe(name string) (int, error) {
 	return bpf.Load(name, C.BPF_PROG_TYPE_KPROBE)
 }
 
@@ -168,6 +181,7 @@ func (bpf *Module) load(name string, progType int) (int, error) {
 }
 
 var kprobeRegexp = regexp.MustCompile("[+.]")
+var uprobeRegexp = regexp.MustCompile("[^a-zA-Z0-9_]")
 
 func (bpf *Module) attachProbe(evName string, attachType uint32, fnName string, fd int) error {
 	if _, ok := bpf.kprobes[evName]; ok {
@@ -187,6 +201,20 @@ func (bpf *Module) attachProbe(evName string, attachType uint32, fnName string, 
 	return nil
 }
 
+func (bpf *Module) attachUProbe(evName string, attachType uint32, path string, addr uint64, fd, pid int) error {
+	evNameCS := C.CString(evName)
+	binaryPathCS := C.CString(path)
+	res, err := C.bpf_attach_uprobe(C.int(fd), attachType, evNameCS, binaryPathCS, (C.uint64_t)(addr), (C.pid_t)(pid), 0, -1, nil, nil)
+	C.free(unsafe.Pointer(evNameCS))
+	C.free(unsafe.Pointer(binaryPathCS))
+
+	if res == nil {
+		return fmt.Errorf("failed to attach BPF kprobe: %v", err)
+	}
+	bpf.uprobes[evName] = res
+	return nil
+}
+
 // AttachKprobe attaches a kprobe fd to a function.
 func (bpf *Module) AttachKprobe(fnName string, fd int) error {
 	evName := "p_" + kprobeRegexp.ReplaceAllString(fnName, "_")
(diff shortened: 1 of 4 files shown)


alban:
Great!

For the string/hexa thing, it seems that bcc has some automatic introspection mechanism via `C.bpf_table_key_snprintf()` to see how to print the key according to its type but I don't know how it works.

louism517:
Sorry for delay, been uber-busy. Readline example to follow.

louism517:
@schu all requested changes have been implemented. Cheers.

schu:
Thanks @louism517 - we have been busy as well :)

I meant to drop the drop the `strlen_count` example app entirely in favor of `bash_readline` (to avoid the adventurous string conversion + lib problem with `strlen_count`) but YMMV.

Can you squash the commits into one with a clean message? Otherwise ok for merge I think :+1: 

@alban please take a look.

bcc/module.go
```diff
@@ -201,6 +229,74 @@ func (bpf *Module) AttachKretprobe(fnName string, fd int) error {
 	return bpf.attachProbe(evName, BPF_PROBE_RETURN, fnName, fd)
 }
 
+// AttachUprobe attaches a uprobe fd to the symbol in the library or binary 'name'
+// The 'name' argument can be given as either a full library path (/usr/lib/..),
+// a library without the lib prefix, or as a binary with full path (/bin/bash)
+// A pid can be given to attach to, or -1 to attach to all processes
+func (bpf *Module) AttachUprobe(name, symbol string, fd, pid int) error {
+	path, addr, err := resolveSymbolPath(name, symbol, 0x0, 0)
```
alban: Does it work when attaching a uprobe to a process in a container?

This causes a call to `C.bcc_resolve_symname()` with  the parameter `pid=0`. bcc temporarily changes the current mount namespace to the mount namespace of the specified pid (if specified, i.e. !=0) in order to read the symbols from the correct files in the correct rootfs when reading `/proc/<pid>/maps`.

Since the pid is not specified here, I suspect `AttachUprobe` does not work correctly when the traced process lives in a different mount namespace than the tracer (like with containers, or with daemons started by systemd with [RootDirectory=](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#RootDirectory=)).

Unfortunately, I think just passing the pid in parameter would not work because "man 2 setns" says:
> A process may not be reassociated with a new mount namespace if it is multithreaded.

And Go programs are always multithreaded.

rkt works around a similar problem by [forking a new single-threaded process to do the setns](https://github.com/rkt/rkt/blob/cb25d9062afdc7ca4fe7d1ebd2c106d2fcee36ed/stage1/app_add/app_add.go#L223-L226).

If this is correct, I suggest to add a comment about this or file a follow-up issue..

bcc/module.go
```diff
@@ -201,6 +229,74 @@ func (bpf *Module) AttachKretprobe(fnName string, fd int) error {
...
+	}
+	return nil
+}
+
+// AttachUretprobe attaches a uretprobe fd to the symbol in the library or binary 'name'
+// The 'name' argument can be given as either a full library path (/usr/lib/..),
+// a library without the lib prefix, or as a binary with full path (/bin/bash)
+// A pid can be given to attach to, or -1 to attach to all processes
+func (bpf *Module) AttachUretprobe(name, symbol string, fd, pid int) error {
+	path, addr, err := resolveSymbolPath(name, symbol, 0x0, 0)
```
alban: ditto

bcc/symbol.go
```diff
@@ -0,0 +1,126 @@
...
+
+	symbol := &bccSymbol{}
+	symbolC := (*C.struct_bcc_symbol)(unsafe.Pointer(symbol))
+	moduleCS := C.CString(module)
+	defer C.free(unsafe.Pointer(moduleCS))
+	symnameCS := C.CString(symname)
+	defer C.free(unsafe.Pointer(symnameCS))
+
+	if pid == -1 {
+		pid = 0
```
alban: do you pass pid == -1 anywhere?

bcc/symbol.go
```diff
@@ -0,0 +1,126 @@
...
+type bccSymbol struct {
+	name         *C.char
+	demangleName *C.char
+	module       *C.char
+	offset       C.ulonglong
+}
+
+// resolveSymbolPath returns the file and offset to locate symname in module
+func resolveSymbolPath(module string, symname string, addr int64, pid int) (string, int64, error) {
+
```
alban: empty line

bcc/symbol.go
```diff
@@ -0,0 +1,126 @@
...
+#include <bcc/bpf_common.h>
+#include <bcc/libbpf.h>
+#include <bcc/bcc_syms.h>
+extern void foreach_symbol_callback(char*, uint64_t);
+*/
+import "C"
+
+type symbolAddress struct {
+	name string
+	addr int64
```
alban: shouldn't it be `uint64`? Same below in this file

examples/bash_readline.go
```diff
@@ -0,0 +1,107 @@
+// Copyright 2016 Kinvolk
```
alban: you can use your own name and/or company name if you wrote the code.

Also, s/2016/2017.

alban:
@louism517 Sorry for the delay! I added my review.

bcc/symbol.go
louism517: Yes, in module.go, in the various AttachUProbe functions, a pid of -1 signifies we wish to attach to all processes (the -1 eventually gets passed through to bpf_attach_uprobe). For e.g.:

```
// AttachUprobe attaches a uprobe fd to the symbol in the library or binary 'name'
// The 'name' argument can be given as either a full library path (/usr/lib/..),
// a library without the lib prefix, or as a binary with full path (/bin/bash)
// A pid can be given to attach to, or -1 to attach to all processes
func (bpf *Module) AttachUprobe(name, symbol string, fd, pid int) error {
	path, addr, err := resolveSymbolPath(name, symbol, 0x0, 0)
	if err != nil {
		return err
	}
	evName := fmt.Sprintf("p_%s_0x%x", uprobeRegexp.ReplaceAllString(path, "_"), addr)
	return bpf.attachUProbe(evName, BPF_PROBE_ENTRY, path, addr, fd, pid)
}
```

So we just switch it to be 0 here, meaning the symbol resolution will not be tied to any specific PID (I think).

bcc/module.go
louism517: Yes, you are absolutely correct - AttachUProbe doesn't work when the traced process is running in a different namespace.

But I'm not entirely sure on the reason.

The call to bcc_resolve_symname() (_with_ the pid) _appears_ to come back correctly, with what looks like the correct address (its a little hard to tell as I'm using an Ubuntu container and the address is the same in the parent and child namespace).

However, the trace itself simply doesn't work, no events are triggered.

I looked through the bcc code and failed to find where it does the namespace switching, would you be able to point me towards it?

If its OK with you I'll add a comment and then raise a follow up issue once the PR is merged?

I have amended the code to actually pass in the pid anyway, as it should.

bcc/symbol.go
louism517: Yes, you're right. Will change throughout.

louism517:
@alban - changes made as requested.

schu:
@louism517 thanks for the fixups!

We have changed `examples` and each tool has its own sub directory now (https://github.com/iovisor/gobpf/tree/master/examples/bcc). Could you move `examples/strlen_count.go` to `examples/bcc/strlen_count/strlen_count.go` and `examples/bash_readline.go` to `examples/bcc/bash_readline/bash_readline.go`. Also, please squash everything into a single commit `bcc: add support for uprobes`. Happy to finally merge after :+1:

louism517:
@schu Changes as requested, and squashed commits. Thanks!

schu:
Thank you :sparkles:

PR #39, opened 2017-03-31
kretprobes: set maxactive

alban:
We are discussing adding a way to specify the maxactive field of a
kretprobe in the kernel:
https://github.com/iovisor/bcc/issues/1072#issuecomment-289777787

How do we expose this in gobpf? I see two possible ways:
- at build-time: `SEC("kretprobe?maxactive=128/inet_csk_accept")`
- at run-time: `m.EnableKprobe("kretprobe/inet_csk_accept", 128)`

This patch implements it at run-time. This might not not ideal because `EnableKprobes` (plural) would use the same maxactive for all kretprobes.

/cc @iaguis @schu

TODO:
- [x] test it

### DIFF
--- a/elf/module.go
+++ b/elf/module.go
@@ -145,7 +145,7 @@ func (b *Module) EnableKprobe(secName string, maxactive int) error {
 	}
 	defer f.Close()
 
-	cmd := fmt.Sprintf("%s%d:%s %s\n", probeType, maxactiveStr, eventName, funcName)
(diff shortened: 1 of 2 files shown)


iaguis:
With this solution, users can do

```
for kp := range m.IterKprobes() {
	if err := m.EnableKprobe(kp.Name, 16); err != nil {
		return nil, err
	}
}
```

iaguis:
Updated.

Tested on a kernel without support for setting maxactive and it works.

iaguis:
Tested on a kernel with support for setting maxactive and it works as expected.

PR #49, opened 2017-05-15
elf: support using pinned maps

schu:


elf/elf.go
```diff
@@ -257,7 +322,21 @@ func elfReadMaps(file *elf.File) (map[string]*Map, error) {
 			mapCount := len(data) / C.sizeof_struct_bpf_map_def
 			for i := 0; i < mapCount; i++ {
 				pos := i * C.sizeof_struct_bpf_map_def
-				cm, err := C.bpf_load_map((*C.bpf_map_def)(unsafe.Pointer(&data[pos])))
+				mapDef := (*C.bpf_map_def)(unsafe.Pointer(&data[pos]))
+
+				var mapPathC *C.char
+				if mapDef.pinning > 0 {
+					mapPath, err := prepareBPFFS(C.GoString(&mapDef.namespace[0]), name)
```
alban: You could verify somewhere that `&mapDef.namespace[0]` is really NULL-terminated before the end of the 256-sized buffer, in case someone builds a ELF file with a corrupted `mapDef.namespace`.

Also, you could check that the namespace:
- does not contain any `/ character
- is not empty (equals to "")
- is not equal to "." or ".."

elf/elf.go
```diff
@@ -257,7 +322,21 @@ func elfReadMaps(file *elf.File) (map[string]*Map, error) {
 			mapCount := len(data) / C.sizeof_struct_bpf_map_def
```
alban: Could you check if `len(data)` is really a multiple of `C.sizeof_struct_bpf_map_def`? If another user of the gobpf library has ELF files with the definition of `struct bpf_map_def` before this PR and use gobpf code after this PR, this would help them to detect the issue.

alban: Actually, since we use the map name included in the name of the ELF section (e.g. `SEC("maps/tcp_event_ipv4")`), we cannot have several `struct bpf_map_def` in one ELF section because we would not know how to distinguish their names.

The idea to put several map definitions in the same ELF section comes from iproute2:
https://git.kernel.org/pub/scm/linux/kernel/git/shemminger/iproute2.git/tree/lib/bpf.c?h=v4.11.0#n1637
They have a single section "maps" for all the maps:
```
		if (data.sec_hdr.sh_type == SHT_PROGBITS &&
		    !strcmp(data.sec_name, ELF_SECTION_MAPS))
			ret = bpf_fetch_maps_begin(ctx, i, &data);
```
They take the map name from the symbol name in the symbol table instead of using the section name like gobpf.

gobpf should either do the same, or stop pretending to support several map definitions in the same ELF section (and check that `len(data) == C.sizeof_struct_bpf_map_def` in the latter case).

alban: For the record, the idea to have one ELF section per map comes from:
https://github.com/safchain/goebpf/blob/b1e37d587ca2eb6a589c72efd7d8306372d3e491/goebpf.go#L469
It's easier because then it does not need to get the name from the symbol table.

schu: > Could you check if `len(data)` is really a multiple of `C.sizeof_struct_bpf_map_def`?

Good point. Otherwise users get a misleading `relocation error` later as the map cannot be found in `b.maps`.

> Actually, since we use the map name included in the name of the ELF section (e.g. `SEC("maps/tcp_event_ipv4")`), we cannot have several `struct bpf_map_def` in one ELF section because we would not know how to distinguish their names.

I'll remove support for several map definitions per `SEC`. We can add it back later based using the symbol name.

elf/elf.go
schu: 9f73136 - please take a look.

alban: 9f73136 lgtm

PR #51, opened 2017-05-18
elf: add `Close()`

schu:


elf/elf.go
```diff
@@ -460,6 +460,7 @@ func (b *Module) Load(parameters map[string]SectionParams) error {
 	if err != nil {
 		return err
 	}
+	defer b.file.Close()
```
alban: https://golang.org/pkg/debug/elf/#File.Close

> If the File was created using NewFile directly instead of Open, Close has no effect.

The file was created with `elf.NewFile`, so it does not seem the correct way to close it.

iaguis: Right, that's like a virtual in-memory File, so I don't think we need to close it.

schu: Thanks for pointing out. Dropped the commit.

schu:
c520ae5 looks good to me.

PR #54, opened 2017-05-25
Fix compilation error in 32-bit architectures

2opremio:
I get error `bpffs/fs.go:20: constant 3405662737 overflows int32` in ARM, see https://circleci.com/gh/weaveworks/scope/7239

### DIFF
--- a/bpffs/fs.go
+++ b/bpffs/fs.go
@@ -3,11 +3,20 @@ package bpffs
 import (
 	"fmt"
 	"syscall"
+	"unsafe"
 )
 
-const (
-	BPFFSPath = "/sys/fs/bpf"
-)
+const BPFFSPath = "/sys/fs/bpf"
+
+var FsMagicBPFFS int32
+
+func init() {
+	// https://github.com/coreutils/coreutils/blob/v8.27/src/stat.c#L275
+	magic := uint32(0xCAFE4A11)
+	// 0xCAFE4A11 overflows an int32, which is what's expected by Statfs_t.Type in 32bit platforms.
+	// To avoid conditional compilation for all 32bit/64bit platforms, we use an unsafe cast
+	FsMagicBPFFS = *(*int32)(unsafe.Pointer(&magic))
+}
 
 // IsMounted checks if the BPF fs is mounted already
 func IsMounted() (bool, error) {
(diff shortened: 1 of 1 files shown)


2opremio:
This is not good enough, I need to rethink this.

bpffs/fs.go
```diff
@@ -8,7 +8,7 @@ import (
 const (
 	BPFFSPath = "/sys/fs/bpf"
 	// https://github.com/coreutils/coreutils/blob/v8.27/src/stat.c#L275
-	FsMagicBPFFS = 0xCAFE4A11
+	FsMagicBPFFS = int64(0xCAFE4A11)
```
alban: `0xCAFE4A11` fits on a `uint32` but not on a `int32`.

`syscall.Statfs_t.Type` can be a `int32 or a `int64` depending on the architecture.

I was wondering if `FsMagicBPFFS` could be declared as a constant of type `syscall.Statfs_t.Type` (which would be substituted to `int32` or `int64`) but I don't think that's possible with Go constants.

And 0xCAFE4A11 would need to be written as a negative value (−889304559) when it's `int32`, but a different negative value when it's `int64` (expending the signed bit `0xffffffff` on the most significant bits).

So I can only think of writing different implementations of `func IsMounted()` for different architectures.

alban: Btw, the canonical definition of the constant is in:
https://github.com/torvalds/linux/blob/v4.8/include/uapi/linux/magic.h#L80
```
#define BPF_FS_MAGIC		0xcafe4a11
```

2opremio: I have solved this by using a literal instead of a constant.

2opremio:
@alban PTAL

2opremio:
OK, that doesn't work either (I wasn't cross-compiling properly when testing)

bpffs/fs.go
2opremio: That doesn't work (I wasn't testing properly)

2opremio:
How about an unsafe cast? Do you think that's acceptable? Otherwise I will cave in and create two different files just to define `FsMagicBPFFS` :)

alban:
Something like this?
```
var FsMagicBPFFS int32

func init() {
  // foo cannot be a constant because we take the address
  foo := uint32(0xCAFE4A11)
  FsMagicBPFFS = *(*int32)(unsafe.Pointer(&foo))
}

...

if int32(data.Type) == FsMagicBPFFS
```

All Linux filesystem magics seem to fit on 32 bits, so it should be ok.

2opremio:
Yep, that's what I meant.

@alban I think I finally got it right. Sorry for the noise

alban:
Could you add the link to https://github.com/torvalds/linux/blob/v4.8/include/uapi/linux/magic.h#L80 while you are at it?
After that, LGTM.

2opremio:
Done, please merge when ready

Issue #35, opened 2017-03-22
bcc: perf map initialization by iterating over cpus fails


alban:
The example in `examples/perf.go` does not work on my laptop:
```
perf_event_open: No such device
   (check your kernel for PERF_COUNT_SW_BPF_OUTPUT support, 4.4 or newer)
Failed to init perf map: failed to open perf buffer: no such device
```

It works on @iaguis' laptop. The difference is that I have more "possible" cpus:
- /sys/devices/system/cpu/possible: 0-7
- /sys/devices/system/cpu/present: 0-3
- /sys/devices/system/cpu/online: 0-3

The python code iterates over cpus to initialize the perf map with [get_online_cpus](https://github.com/iovisor/bcc/blob/aaab74eb2633745e58ba421ba163b56f7b74f035/src/python/bcc/table.py#L523). The go code iterates with [bpf_get_next_key](https://github.com/iovisor/gobpf/blob/db0602ef0f04700167819a555d40ac49cad99494/bcc/perf.go#L142)

```
[pid   680] perf_event_open({type=PERF_TYPE_SOFTWARE, size=0 /* PERF_ATTR_SIZE_??? */, config=PERF_COUNT_SW_BPF_OUTPUT, ...}, -1, 0, -1, PERF_FLAG_FD_CLOEXEC) = 9
[pid   680] bpf(BPF_MAP_UPDATE_ELEM, {map_fd=3, key=0xc208000920, value=0xc208000924, flags=BPF_ANY}, 48) = 0
[pid   680] bpf(BPF_MAP_GET_NEXT_KEY, {map_fd=3, key=0xc208000920, next_key=0xc208000920}, 48) = 0
[pid   680] perf_event_open({type=PERF_TYPE_SOFTWARE, size=0 /* PERF_ATTR_SIZE_??? */, config=PERF_COUNT_SW_BPF_OUTPUT, ...}, -1, 1, -1, PERF_FLAG_FD_CLOEXEC) = 10
[pid   680] bpf(BPF_MAP_UPDATE_ELEM, {map_fd=3, key=0xc208000920, value=0xc208000924, flags=BPF_ANY}, 48) = 0
[pid   680] bpf(BPF_MAP_GET_NEXT_KEY, {map_fd=3, key=0xc208000920, next_key=0xc208000920}, 48) = 0
[...]

PR #55, opened 2017-05-28
Fix bpf_open_perf_buffer call after commit  4b764de6bcb3655b29b2fb6df6c9c65b7a31338c

wofanli:
BCC commit 4b764de6bcb3655b29b2fb6df6c9c65b7a31338c introduces callback for perf buffer lost sample events. 
Need to adjust bpf_open_perf_buffer calling in gobpf.
@schu without the adjustment, we might meet the same error log with #35.

### DIFF
--- a/bcc/perf.go
+++ b/bcc/perf.go
@@ -123,7 +123,7 @@ func InitPerfMap(table *Table, receiverChan chan []byte) (*PerfMap, error) {
 	cpu := 0
 	res := 0
 	for res == 0 {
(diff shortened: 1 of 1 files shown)


iaguis:
The callback introduced in https://github.com/iovisor/bcc/commit/4b764de6 is not yet in any bcc release, so this will break using gobpf with the latest bcc release (0.3.0).

I'm not sure how to handle this, maybe we should have a bcc-0.3.0 branch to keep compatibility and try to keep master compatible with latest bcc master.

wofanli:
Agree to keep master compatible with latest bcc master and keep the release version aligned with bcc, which would be clear for gobpf user.

schu:
Closing in favor of #73

PR #57, opened 2017-05-29
elf: add socket filter support

iaguis:
We follow kernel's convention of taking sections that start with
"socket".

Also, rework cgroup API for consistency.

### DIFF
--- a/elf/module.go
+++ b/elf/module.go
@@ -63,8 +63,19 @@ int bpf_prog_attach(int prog_fd, int target_fd, enum bpf_attach_type type)
 	attr.attach_bpf_fd = prog_fd;
 	attr.attach_type   = type;
 
-	// BPF_PROG_ATTACH = 8
-	return syscall(__NR_bpf, 8, &attr, sizeof(attr));
+	return syscall(__NR_bpf, BPF_PROG_ATTACH, &attr, sizeof(attr));
+}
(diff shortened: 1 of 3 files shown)


elf/module.go
```diff
@@ -66,6 +66,19 @@ int bpf_prog_attach(int prog_fd, int target_fd, enum bpf_attach_type type)
...
+{
+	union bpf_attr attr;
+
+	bzero(&attr, sizeof(attr));
+	attr.target_fd	   = target_fd;
+	attr.attach_bpf_fd = prog_fd;
+	attr.attach_type   = type;
+
+	// BPF_PROG_DETACH = 9
+	return syscall(__NR_bpf, 9, &attr, sizeof(attr));
```
schu: We have `BPF_PROG_ATTACH` / `BPF_PROG_DETACH` defined in `elf/include/bpf.h` - any reason we don't use them?

schu:
Could you add `socket` to the supported elf sections in the README? Otherwise looks good to me!

elf/module.go
iaguis: Good point, I didn't see that 🤔

iaguis:
Updated.

schu:
`module_unsupported.go` needs an update.

PR #62, opened 2017-06-14
elf: don't close event FD if Kprobe wasn't enabled

iaguis:
Trying to close it was making `Module.Close()` fail for Kprobes that
weren't enabled.

### DIFF
--- a/elf/elf.go
+++ b/elf/elf.go
@@ -562,6 +562,7 @@ func (b *Module) Load(parameters map[string]SectionParams) error {
 						Name:  secName,
 						insns: insns,
 						fd:    int(progFd),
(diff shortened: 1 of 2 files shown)


iaguis:
Changed PR since it seems we don't even need to keep that fd open after EnableKprobe. Thanks for the suggestion @alban.

iaguis:
Wait, I think I was testing the wrong thing. Stay tuned...

iaguis:
> Changed PR since it seems we don't even need to keep that fd open after EnableKprobe. Thanks for the suggestion @alban.

Unfortunately this doesn't work. I don't get any events if I do it.

PR #64, opened 2017-06-29
elf: add `CloseExt`

schu:
Pinned maps now need to be unpinned explicitly by setting `Unpin` in the
`CloseOptions` and using `CloseExt`.

### DIFF
--- a/elf/module.go
+++ b/elf/module.go
@@ -439,9 +439,10 @@ func unpinMap(m *Map) error {
 	return syscall.Unlink(mapPath)
 }
 
-func (b *Module) closeMaps() error {
+func (b *Module) closeMaps(options map[string]CloseOptions) error {
 	for _, m := range b.maps {
-		if m.m.def.pinning > 0 {
+		doUnpin := options[fmt.Sprintf("maps/%s", m.Name)].Unpin
+		if m.m.def.pinning > 0 && doUnpin {
 			unpinMap(m)
 		}
 		for _, fd := range m.pmuFDs {
@@ -457,6 +458,12 @@ func (b *Module) closeMaps() error {
 	return nil
 }
 
+// CloseOptions can be used for custom `Close` parameters
+type CloseOptions struct {
+	// Set Unpin to true to close pinned maps as well
+	Unpin bool
+}
+
 // Close takes care of terminating all underlying BPF programs and structures.
 // That is:
 //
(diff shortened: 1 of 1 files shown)


elf/module.go
```diff
@@ -467,8 +472,14 @@ func (b *Module) closeMaps() error {
 //
 // It doesn't detach BPF programs from cgroups or sockets because they're
 // considered resources the user controls.
+// It also doesn't unpin pinned maps. Use CloseExt and set Unpin to do this.
 func (b *Module) Close() error {
-	if err := b.closeMaps(); err != nil {
+	return b.CloseExt(nil)
+}
+
+// CloseExt takes a map "elf section -> CloseOptions"
```
alban: it should be documented somewhere whether it unpins a map if it is not mentioned in the map. What's the default value of the Unpin bool?

elf/module.go
```diff
@@ -457,6 +458,10 @@ func (b *Module) closeMaps() error {
 	return nil
 }
 
+type CloseOptions struct {
```
alban: Exported types should be documented, I think it is requested by `go fmt` or `lint`.

elf/module.go
iaguis: the default value of bools is false

elf/module.go
schu: golint it is. There's more stuff it complains about that I'll do in a separate PR. Thanks!

elf/module.go
```diff
@@ -457,6 +458,12 @@ func (b *Module) closeMaps() error {
 	return nil
 }
 
+// CloseOptions can be used for custom `Close` parameters
+type CloseOptions struct {
+	// Set Unpin to true to close pinned maps as well
+	Unpin bool
```
iaguis: Isn't the convention that struct fields are commented inline?

elf/module.go
iaguis: Now that I see this, it's kinda confusing, because it takes a map of elf map sections. Wdyt about:

```
CloseExt takes a map "elf map section -> CloseOptions".
```

Also, can you add:

```
If an elf map section is not mentioned in the map, it won't be unpinned.
```

elf/module.go
schu: I don't think so. See e.g. https://github.com/golang/go/blob/master/src/net/http/server.go#L225

iaguis: ok, I guess it doesn't matter and it's done only for small comments (like https://golang.org/src/go/doc/example.go) :+1:

elf/module.go
schu: That we use/consider `CloseOptions` and `Unpin` only for maps could change in the future; would you prefer to change it to `elf map section` nonetheless?

iaguis: Fair enough, let's keep it the way it is now.

PR #71, opened 2017-08-16
perf: close channels on the sender side & fix closing race

alban:
Go channels should always be closed on the sender side. Also, fix the
comment that says otherwise.

When synchronously sending an event, check if it's time to close instead
of blocking indefinitely.

-----

/cc @schu @iaguis @2opremio

### DIFF
--- a/elf/perf.go
+++ b/elf/perf.go
@@ -125,6 +125,9 @@ func InitPerfMap(b *Module, mapName string, receiverChan chan []byte, lostChan c
 	if !ok {
 		return nil, fmt.Errorf("no map with name %s", mapName)
 	}
+	if receiverChan == nil {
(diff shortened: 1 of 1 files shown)


elf/perf.go
```diff
@@ -163,6 +163,11 @@ func (pm *PerfMap) PollStart() {
 		pageSize := os.Getpagesize()
 		state := C.struct_read_state{}
 
+		defer func() {
+			close(pm.receiverChan)
+			close(pm.lostChan)
```
schu: This does need a `if pm.lostChan != nil`, I think?

alban: Right. It works for tcptracer-bpf but I forgot about the generic case. I'll fix it.

PR #72, opened 2017-08-22
elf: object pinning with custom configs

dongsupark:
Introduce object pinning with custom configs. When a module is loaded with parameters of custom config, pin the object with a path defined in the custom config. The actual pinnned object is visible as `/sys/fs/bpf/<pin_path>`. The pin_path can be any path, e.g. a normal path like `ns1/testgroup1`, or hash of a particular path.

To cover the test case of loading modules with custom pinning config, also add a test for loading module with specific parameters, and checking for actual pinned object under `/sys/fs/bpf`.


elf/module.go
```diff
@@ -143,24 +144,48 @@ type TracepointProgram struct {
 	efd   int
 }
 
+// CustomContext represents a custom context
+type CustomContext struct {
+	Name       string
+	CgroupPath string
```
alban: It should not be specific to cgroups, but follow similar pattern to `/etc/iproute2/bpf_pinning`, like a `map[int]string`.

schu:
From a cursory look:

`PinObjectCustom`  should allow the user to pin an object to a custom path, but it doesn't do that atm. I.e. the hashing of `customPath` is not something that should be in gobpf.

We have a `SectionParams` map that can be passed to `Load()` to set individual parameters for sections in an .elf file:

https://github.com/iovisor/gobpf/blob/master/elf/elf.go#L442-L449

Did you try to add a new field to `SectionParams` instead? For instance

```
type SectionParams struct {
	PerfRingBufferPageCount   int
	SkipPerfMapInitialization bool
        PinConfig MapPinConfig
}
```

(With a better name than `PinConfig` maybe :)

`func NewModuleWithCustomCtx(fileName string, customCtx map[string]*CustomContext) *Module` doesn't feel right IMHO:

* `context` or `ctx` in Go land usually is associated with https://godoc.org/context
* While the argument `customCtx` sounds very generic it serves only a very specific purpose at the same time

elf/pinning.go
```diff
@@ -67,3 +76,24 @@ func pinObject(fd int, namespace, object, name string) error {
...
+		return fmt.Errorf("error checking if %q is mounted: %v", BPFFSPath, err)
+	}
+	if !mounted {
+		return fmt.Errorf("bpf fs not mounted at %q", BPFFSPath)
+	}
+
+	if customPath == "" {
+		return fmt.Errorf("no custom path is given")
+	}
+	customHash := hex.EncodeToString(util.MakeSha256([]byte(customPath)))
```
alban: I think it should not be up to gobpf to decide whether the path should be hashed or not. The caller of the gobpf library should decide that. Some users of gobpf might use cgroup path, but other users (such as iproute2's tc) might not.

alban:
@schu good point about `Load(parameters)`, I agree that sounds better than hooking the new feature on `NewModule()`

bpf_test.go
```diff
@@ -266,4 +303,37 @@ func TestModuleLoadELF(t *testing.T) {
...
+	})
+	if bc == nil {
+		t.Fatal("prog is nil")
+	}
+	if err := bc.Load(nil); err != nil {
+		if !strings.Contains(err.Error(), "invalid argument") {
+			t.Fatal(err)
+		}
+	}
+	defer bc.Close()
```
nhlfr: Shoudn't this line go before `bc.Load(nil)`

dongsupark: @nhlfr Right. This is a result of simple copy-and-paste. In the same function there's another place where `bc.Close()` is called after `bc.Load()`... ;-)
Will fix it later.

iaguis: Actually this is ok, `Close()` reverts the effects of `Load()`. `NewModule()` just allocates some maps and returns a `Module` object

alban: Do you think the documentation on Close() could be improved to say that?
https://github.com/iovisor/gobpf/blob/36406ec/elf/module.go#L540-L551

iaguis: How about?

```
Close takes care of terminating all underlying BPF programs and structures instantiated by Load()
```

bpf_test.go
```diff
@@ -234,6 +239,26 @@ func checkSocketFilters(t *testing.T, b *elf.Module) {
 	}
 }
 
+func checkPinCustomConfig(t *testing.T, b *elf.Module) {
+	// NOTE: this test doesn't really work kernel 4.9 or older,
+	// because /sys/fs/bpf/fw doesn't exist at all.
```
alban: What is the name of the feature that is missing before 4.9?

https://github.com/iovisor/bcc/blob/master/docs/kernel-versions.md

alban: did you `mkdir` the directory where you want to pin a bpf object before?

elf/elf.go
```diff
@@ -262,7 +262,16 @@ static int perf_event_open_map(int pid, int cpu, int group_fd, unsigned long fla
 */
 import "C"
 
-const useCurrentKernelVersion = 0xFFFFFFFE
+const (
+	useCurrentKernelVersion = 0xFFFFFFFE
+	PIN_NONE                = 0
+	PIN_OBJECT_NS           = 1
+	PIN_GLOBAL_NS           = 2
```
alban: could you add a link to iproute2/tc where those values are defined / explained?

bpf_test.go
dongsupark: @alban I'm wondering too which one exactly causes the failure. Just looking at the wiki, it isn't clear to me.

pkg/util/util.go
```diff
@@ -0,0 +1,23 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package util
+
+import "crypto/sha256"
+
+func MakeSha256(data []byte) []byte {
```
alban: This is now only used for the tests. This should be removed. We can do tests without hashing.

elf/elf.go
```diff
@@ -442,6 +451,18 @@ func (b *Module) relocate(data []byte, rdata []byte) error {
 type SectionParams struct {
 	PerfRingBufferPageCount   int
 	SkipPerfMapInitialization bool
+	CustomConfig              []PinCustomConfig
```
schu: `Custom Config` should be named differently I think, e.g. `PinConfig`. It's a very generic name while the field is solely for pinning.

elf/elf.go
```diff
@@ -442,6 +451,18 @@ func (b *Module) relocate(data []byte, rdata []byte) error {
...
+}
+
+// PinCustomConfig represents a custom config for pinning.
+// Id has to be greater than 2, as numbers 0,1,2 are reserved.
+// The final path will be composed in a form of "/sys/fs/bpf/PinPath",
+// where "/sys/fs/bpf" is a mount point of the standard bpf filesystem.
+// PinPath can be any path, e.g. a normal path like "ns1/testgroup1",
+// or sha256 string of a particular path.
+type PinCustomConfig struct {
+	Id      uint32 // >= 3
```
schu: `Id` doesn't seem to have any actual purpose? Do I miss something?

elf/elf.go
```diff
@@ -674,6 +695,17 @@ func (b *Module) Load(parameters map[string]SectionParams) error {
 					fd:    int(progFd),
 				}
 			}
+
+			if params, ok := parameters[secName]; ok && params.CustomConfig != nil {
```
schu: Further above we load programs from `elf.SHT_REL` - can those not be pinned?

schu: Even if not in the scope for this PR, we should think about how maps can be pinned to a custom path as well.

elf/elf.go
schu: Also, I'm not sure if we should have / need a slice of `PinCustomConfig`. For example, what if we add more fields later and those have conflicts?

bpf_test.go
```diff
@@ -266,4 +292,47 @@ func TestModuleLoadELF(t *testing.T) {
...
+	}
+
+	var secParams = map[string]elf.SectionParams{
+		"cgroup/skb": elf.SectionParams{
+			PerfRingBufferPageCount:   0,
+			SkipPerfMapInitialization: false,
+			CustomConfig: []elf.PinCustomConfig{
+				elf.PinCustomConfig{
+					Id:      3,
+					PinPath: filepath.Join("fw", hex.EncodeToString(util.MakeSha256([]byte("testgroup1")))),
```
schu: The only difference to `testgroup2` seems to be that `util.MakeSha256` and `hex.EncodeToString` is used, i.e. it doesn't seem to test a different code path and is redundant? Do I miss something?

elf/pinning.go
```diff
@@ -45,7 +45,14 @@ func pinObject(fd int, namespace, object, name string) error {
...
 	}
 	mapPath := filepath.Join(BPFFSPath, namespace, object, name)
-	err = os.MkdirAll(filepath.Dir(mapPath), 0755)
+	if err := pinObjectCommon(fd, mapPath); err != nil {
+		return err
+	}
+	return nil
+}
+
+func pinObjectCommon(fd int, mapPath string) error {
```
schu: Can `PinObjectGlobal` maybe use `PinObjectCustom`, i.e. is `pinObject` still needed?

bpf_test.go
dongsupark: @schu I'll just remove the sha256 test, as @alban suggested. So we won't need to test twice at all.

elf/elf.go
alban: It has to match the `pinning` value in `struct bpf_map_def`:
https://github.com/iovisor/gobpf/blob/09e8636/elf/elf.go#L68

One bpf module can have several maps, and each map can have a different `pinning` field. So we need to be able to pass several `PinCustomConfig` configuration in the `Load()` to define potentially several mappings.

schu: > It has to match the `pinning` value in struct `bpf_map_def`

Hmm, I don't see the purpose,  seem redundant in the case of maps and not needed here, since the PR doesn't consider custom pinning for maps during `Load()`, if I haven't missed something.

> One bpf module can have several maps, and each map can have a different pinning field.

But we can have only one map per section, and the parameters are per section (`SectionParams`)?

elf/elf.go
alban: > Further above we load programs from `elf.SHT_REL` - can those not be pinned?

Yes, they can. Ideally, the code should be refactored: we have duplicate code for bpf programs in SHT_REL and those that don't use symbols (i.e. programs that don't use maps). This duplication comes from the original code on [goebpf](https://github.com/safchain/goebpf/blob/master/goebpf.go).

If it makes things easier, we could focus on pinning bpf maps only and not bpf programs. And do the program pinning in a future PR. In practice, we have use cases for map pinning, but I can't think of a use case for program pinning at this moment. Up to you.

elf/elf.go
```diff
@@ -674,6 +695,17 @@ func (b *Module) Load(parameters map[string]SectionParams) error {
...
+				for _, cc := range params.CustomConfig {
+					if isPinReserved(cc.Id) {
+						continue
+					}
+					if err := PinObjectCustom(int(progFd), cc.PinPath); err != nil {
```
alban: We should pin a bpf object only one time and not pin it several time.

Here you pin a bpf program and not a bpf map. But the custom pinning parameter has to be correlated with the `pinning` field in the bpf_map_def`. We have no equivalent `struct bpf_prog_def` at all at the moment, so we can't decide where to pin a program... I would just not pin bpf programs at all at the moment, as per previous comment.

elf/elf.go
alban: > But we can have only one map per section, and the parameters are per section (`SectionParams`)?

Oh, I missed that...

tests/dummy.c
```diff
@@ -20,6 +20,13 @@ enum bpf_map_type {
 	BPF_MAP_TYPE_CGROUP_ARRAY,
 };
 
+enum bpf_pin_type {
+    PIN_NONE = 0,
+    PIN_OBJECT_NS,
+    PIN_GLOBAL_NS,
+    PIN_CUSTOM_NS,
+};
```
alban: We should have a canonical header file that contains the definition of `bpf_pin_type` and `bpf_map_def` instead of letting projects using gobpf guessing. Then, this test file can `#include` it.

At the moment, `struct bpf_map_def` is defined in `elf/elf.go` and `tests/dummy.c` and the definitions don't match betcause it's difficult to keep things up-to-date with copy/paste.

dongsupark: @alban Updated. Now common structures and types are defined in `elf/include/bpf.h`.

schu:
I think we can further simplify the logic and code by unifying the handling of `PIN_GLOBAL_NS` and `PIN_CUSTOM_NS`. Ultimately, `PIN_GLOBAL_NS` is just a custom path following a convention.

Something like this maybe:

```go
func createPinPath(path string) (string, error) {                                      
        err := bpffs.Mount()               
        if err != nil {                    
                return "", err             
        }                                  
        err = os.MkdirAll(filepath.Dir(path), syscall.S_IRWXU)                         
        if err != nil {                    
                return "", fmt.Errorf("error creating map directory %q: %v", filepath.Dir(path), err)                                                                         
        }                                  
        return path, nil                   
}                                          

func validateMapPath(path string) error {  
        return nil // TODO                 
}                                          

func getMapNamespace(mapDef *C.bpf_map_def) string {                                   
        namespacePtr := &mapDef.namespace[0]                                           
        return C.GoStringN(namespacePtr, C.int(C.strnlen(namespacePtr, C.BUF_SIZE_MAP_NS)))                                                                                   
}                                          

func getMapPath(mapDef *C.bpf_map_def, mapName string, params SectionParams) (string, error) {                                                                                
        var mapPath string                 
        switch mapDef.pinning {            
        case PIN_OBJECT_NS:                
                return "", fmt.Errorf("not implemented yet")                           
        case PIN_GLOBAL_NS:                
                namespace := getMapNamespace(mapDef)                                   
                if namespace == "" {       
                        return "", fmt.Errorf("map %q has empty namespace", mapName)   
                }                          
                mapPath = filepath.Join(BPFFSPath, namespace, "globals", mapName)      
        case PIN_CUSTOM_NS:                   
                if params.PinPath == "" {         
                        return "", fmt.Errorf("no pin path given for map %q with PIN_CUSTOM_NS", mapName)                                                                     
                } 
                mapPath = filepath.Join(BPFFSPath, params.PinPath) 
        default:                           
                // map is not pinned       
                return "", nil             
        }                                  
        if err := validateMapPath(mapPath); err != nil {                               
                return "", err             
        }                                  
        return createPinPath(mapPath)      
} 
```

I feel it would be easier to read and avoids the `ErrNoPinPath` logic.

alban:
We should start having a `Documentation/` directory.

- [ ] explaining what is pinning and the bpffs
- [ ] why it improves the situation compared to the previous solutions in "tc-exec" or SCM_RIGHTS (see description in `man tc-bpf(8)`, search for "can then share eBPF maps with each other" for more context)
- [ ] different pinning options
- [ ] comparisons with other projects doing pinning (iproute2/tc, perf, Linux bpf-samples)
- [ ] when are the map resetted

dongsupark:
@schu Updated. Thanks!

iaguis:
We need to handle `CloseExt()`

dongsupark:
Updated.

* `CloseExt()` already cleaned up maps when Unpin option is set in `closeOptions`. However, `unpinMap()` still cleaned up only global maps. So I had to add another option `PinPath` to `closeOptions`, for unpinning the path correctly for both global and custom namespaces.
* I also updated integration tests, to verify the unpinned path.
* I wrote a simple documentation about object pinning. As discussed, the document doesn't include all other use cases, but is just a simple README.

elf/elf.go
```diff
@@ -245,7 +246,16 @@ static int perf_event_open_map(int pid, int cpu, int group_fd, unsigned long fla
 */
 import "C"
 
-const useCurrentKernelVersion = 0xFFFFFFFE
+const (
+	useCurrentKernelVersion = 0xFFFFFFFE
+
+	// Object pin settings should corresponds to those of other projects, e.g.:
```
schu: `s/correspons/correspond/`

Documentation/pinning.md
```diff
@@ -0,0 +1,59 @@
...
+* `PIN_NONE` : object is not pinned
+* `PIN_OBJECT_NS` : pinning that is local to an object (to-be-implemented)
+* `PIN_GLOBAL_NS` : pinning with a global namespace under e.g. `/sys/fs/bpf/ns1/globals`
+* `PIN_CUSTOM_NS` : pinning with a custom path given as section parameter
+
+### pinning with `PIN_CUSTOM_NS`
+
+When loading a module with `C.bpf_map_def.pinning` set to `PIN_CUSTOM_NS`,
+an additional path must be set to `elf.SectionParams.PinPath` as parameter
+to `Load()`. For example:
```
alban: I would also give an example of the map definition in the `.c` file

Documentation/pinning.md
```diff
@@ -0,0 +1,59 @@
...
+
+```
+bc := elf.NewModule(customELF)
```
alban: what does `bc` stand for?

elf/elf.go
```diff
@@ -276,31 +286,69 @@ func elfReadVersion(file *elf.File) (uint32, error) {
...
+		return "", nil
 	}
 	return mapPath, nil
 }
 
-func validMapNamespace(namespaceRaw *C.char) (string, error) {
-	namespace := C.GoStringN(namespaceRaw, C.int(C.strnlen(namespaceRaw, C.BUF_SIZE_MAP_NS)))
-	if namespace == "" || namespace == "." || namespace == ".." {
-		return "", fmt.Errorf("namespace must not be %q", namespace)
+func getCreateMapPath(mapDef *C.bpf_map_def, mapName string, params SectionParams) (string, error) {
```
schu: Nit: can we give it a different name, maybe just `createMapPath`?

elf/elf.go
```diff
@@ -276,31 +286,69 @@ func elfReadVersion(file *elf.File) (uint32, error) {
...
 	if err != nil {
-		return "", fmt.Errorf("error creating map directory %q: %v", filepath.Dir(mapPath), err)
+		return "", fmt.Errorf("error creating map directory %q: %v", filepath.Dir(path), err)
+	}
+	return path, nil
+}
+
+func validateMapPath(path string) error {
+	relPath := strings.TrimPrefix(path, BPFFSPath)
+	pathsArr := filepath.SplitList(relPath)
```
schu: This does split by `:`, not `/` - intended?

schu: Can you add a unit test for `validateMapPath`?

bpf_test.go
```diff
@@ -263,4 +293,62 @@ func TestModuleLoadELF(t *testing.T) {
...
+		},
+	}
+	defer func() {
+		if err := bc.CloseExt(closeOptions); err != nil {
+			t.Fatal(err)
+		}
+		checkPinConfigCleanup(t, bc)
+	}()
+
+	for _, param := range secParams {
```
schu: Line 344 to 347 (including) can/should be removed, I think.

tests/custom.c
```diff
@@ -0,0 +1,25 @@
...
+struct bpf_map_def SEC("maps/dummy_hash_custom") dummy_hash_custom = {
+	.type = BPF_MAP_TYPE_HASH,
+	.key_size = sizeof(int),
+	.value_size = sizeof(unsigned int),
+	.max_entries = 128,
+	.pinning = PIN_CUSTOM_NS,
+	.namespace = "fw"
+};
+
+struct bpf_map_def SEC("maps/dummy_array_custom") dummy_array_custom = {
```
schu: IMHO it's sufficient to test with one map `PIN_CUSTOM_NS` for gobpf matters and rely on the kernel to work with other maps types as well.

schu: i.e. remove the `testgroup2` test.

bpf_test.go
```diff
@@ -263,4 +293,62 @@ func TestModuleLoadELF(t *testing.T) {
...
+		if err := bpffs.Mount(); err != nil {
+			t.Skipf("error mounting bpf fs, skipping test: %v", err)
+		}
+	}
+
+	var secParams = map[string]elf.SectionParams{
+		"maps/dummy_array_custom": elf.SectionParams{
+			PerfRingBufferPageCount:   0,
+			SkipPerfMapInitialization: false,
+			PinPath:                   filepath.Join("fw", "testgroup1"),
```
schu: Nit: Can you use `gobpf-test` or similar instead of `fw` to avoid confusion?

Documentation/pinning.md
dongsupark: @alban Done. I added a simple map definition to the example.

Documentation/pinning.md
dongsupark: @alban `bc` stands for bpf custom, but in this example, I just changed it to `b`.

Documentation/pinning.md
```diff
@@ -0,0 +1,74 @@
...
+/*
+#include <linux/bpf.h>
+
+struct bpf_map_def SEC("maps/dummy_array_custom") dummy_array_custom = {
+	.type = BPF_MAP_TYPE_ARRAY,
+	.key_size = sizeof(int),
+	.value_size = sizeof(unsigned int),
+	.max_entries = 1024,
+	.pinning = PIN_CUSTOM_NS,
+	.namespace = "ns1"
```
alban: is the `namespace` field used at all when the `PinPath` is specified below?

Documentation/pinning.md
alban: :+1:

Documentation/pinning.md
dongsupark: @alban I'm not sure about that, but without namespace, the map cannot be loaded at all. That seems to be like that when the map definition has its own pinning option.

dongsupark:
@schu Done. Thanks!

Documentation/pinning.md
schu: For me, tests pass without; maybe the error was due to something else?

elf/elf.go
```diff
@@ -293,31 +286,69 @@ func elfReadVersion(file *elf.File) (uint32, error) {
...
+	err = os.MkdirAll(filepath.Dir(path), syscall.S_IRWXU)
 	if err != nil {
-		return "", fmt.Errorf("error creating map directory %q: %v", filepath.Dir(mapPath), err)
+		return "", fmt.Errorf("error creating map directory %q: %v", filepath.Dir(path), err)
+	}
+	return path, nil
+}
+
+func validateMapPath(path string) bool {
+	relPath := strings.TrimPrefix(path, BPFFSPath)
```
schu: Can you also check if the path actually starts with `BPFFSPath` and add a test for that?

bpf_test.go
```diff
@@ -263,4 +291,48 @@ func TestModuleLoadELF(t *testing.T) {
...
+	if !mounted {
+		if err := bpffs.Mount(); err != nil {
+			t.Skipf("error mounting bpf fs, skipping test: %v", err)
+		}
+	}
+
+	var secParams = map[string]elf.SectionParams{
+		"maps/dummy_array_custom": elf.SectionParams{
+			PerfRingBufferPageCount:   0,
+			SkipPerfMapInitialization: false,
```
schu: No need to initialize `PerfRingBufferPageCount` and `SkipPerfMapInitialization`: https://play.golang.org/p/88D3orz3xR

bpf_test.go
```diff
@@ -263,4 +291,48 @@ func TestModuleLoadELF(t *testing.T) {
...
+	bc := elf.NewModule(customELF)
+	if bc == nil {
+		t.Fatal("prog is nil")
+	}
+
+	mounted, err := bpffs.IsMounted()
+	if err != nil {
+		t.Fatalf("error checking if %q is mounted: %v", bpffs.BPFFSPath, err)
+	}
+	if !mounted {
```
schu: `bpffs.Mount()` does this check, so I think we don't need it here?

elf/module.go
```diff
@@ -504,20 +503,22 @@ func (b *Module) closeSocketFilters() error {
...
 }
 
 func (b *Module) closeMaps(options map[string]CloseOptions) error {
 	for _, m := range b.maps {
 		doUnpin := options[fmt.Sprintf("maps/%s", m.Name)].Unpin
 		if m.m.def.pinning > 0 && doUnpin {
-			unpinMap(m)
+			pinPath := options[fmt.Sprintf("maps/%s", m.Name)].PinPath
+			if err := unpinMap(m, pinPath); err != nil {
+				return fmt.Errorf("error unpinning path %q: %v", pinPath, err)
```
schu: This would give a misleading error message in case unpinning of a `PIN_GLOBAL_NS` map fails (w/o a `PinPath`).

elf/elf.go
```diff
@@ -293,31 +286,69 @@ func elfReadVersion(file *elf.File) (uint32, error) {
 	return 0, nil
 }
 
-func prepareBPFFS(namespace, name string) (string, error) {
+func createPinPath(path string) (string, error) {
 	err := bpffs.Mount()
```
schu: Can you put this on one line: `if err := bpffs.Mount(); err != nil {` Same of the next error check below.

Documentation/pinning.md
dongsupark: @schu True. Now I cannot reproduce that error. @.@ I'll remove the namespace from the document.

Documentation/pinning.md
```diff
@@ -0,0 +1,73 @@
...
+struct bpf_map_def SEC("maps/dummy_array_custom") dummy_array_custom = {
+	.type = BPF_MAP_TYPE_ARRAY,
+	.key_size = sizeof(int),
+	.value_size = sizeof(unsigned int),
+	.max_entries = 1024,
+	.pinning = PIN_CUSTOM_NS,
+};
+*/
+
+import "C"
```
alban: Why do we have this `import "C"`? It does not seem to be used in the Go code below. And BPF programs cannot be included in a Go source file in this way, so that's confusing.

Documentation/pinning.md
```diff
@@ -0,0 +1,73 @@
...
+
+b := elf.NewModule(customELF)
```
alban: I would name the variable `customELFFileName` or something, to be clearer that it is not a string containing the .c source code above.

Documentation/pinning.md
```diff
@@ -0,0 +1,73 @@
...
+    "maps/dummy_array_custom": elf.SectionParams{
+        PinPath: "ns1/test1",
+    },
+}
+if err := b.Load(secParams); err != nil {
+    fmt.Println(err)
+}
+
+```
+
```
alban: add a comment to explain how to check if it is pinned.

```
$ ls -l /sys/fs/bpf/ns1/test1
...
```

so we can see how the PinPath correlate with the path on the bpffs.

Documentation/pinning.md
```diff
@@ -0,0 +1,73 @@
...
+    "maps/dummy_array_custom": elf.CloseOptions{
+        Unpin:   true,
+        PinPath: "ns1/test1",
+    },
+}
+if err := b.CloseExt(closeOptions); err != nil {
+    fmt.Println(err)
+}
+
+```
```
alban: you can add it is also possible to use
```
os.Remove("/sys/fs/bpf/ns1/test1");
```

elf/elf.go
```diff
@@ -293,31 +286,71 @@ func elfReadVersion(file *elf.File) (uint32, error) {
...
+
+func validateMapPath(path string) bool {
+	if !strings.HasPrefix(path, BPFFSPath) {
+		return false
+	}
+
+	relPath := strings.TrimPrefix(path, BPFFSPath)
+	pathsArr := strings.Split(relPath, "/")
+	for _, p := range pathsArr {
+		if p == "." || p == ".." {
```
alban: I would do:
```
if filepath.Clean(path) != path {
  return false
}
```

elf/include/bpf.h
```diff
@@ -605,4 +605,28 @@ struct xdp_md {
...
+	unsigned int type;
+	unsigned int key_size;
+	unsigned int value_size;
+	unsigned int max_entries;
+	unsigned int map_flags;
+	unsigned int pinning;
+	char namespace[BUF_SIZE_MAP_NS];
+} bpf_map_def;
+
+typedef struct bpf_map {
```
alban: struct `bpf_map_def` and struct `bpf_map` should not be in the same header file:

- struct bpf_map_def is public, it is #included in the BPF .c source code
- struct bpf_map is an internal type for gobpf that should not be exposed. It is only used in `elf/elf.go`

So I would remove `bpf_map` and put it back in `elf/elf.go`

elf/module.go
```diff
@@ -504,20 +503,29 @@ func (b *Module) closeSocketFilters() error {
...
+		if doUnpin {
+			mapDef := m.m.def
+			if mapDef.pinning == PIN_CUSTOM_NS {
+				pinPath := options[fmt.Sprintf("maps/%s", m.Name)].PinPath
+				if err := unpinMap(m, pinPath); err != nil {
+					return fmt.Errorf("error unpinning path %q: %v", pinPath, err)
+				}
+			} else if mapDef.pinning == PIN_GLOBAL_NS {
+				// ignore error for global objects as pinPath would always be an empty string.
+				unpinMap(m, "")
```
alban: which error do you ignore? `unpinMap()` does not seem to error out on global pinning when PinPath is empty.

elf/module.go
```diff
@@ -504,20 +503,29 @@ func (b *Module) closeSocketFilters() error {
...
 
 func (b *Module) closeMaps(options map[string]CloseOptions) error {
 	for _, m := range b.maps {
 		doUnpin := options[fmt.Sprintf("maps/%s", m.Name)].Unpin
-		if m.m.def.pinning > 0 && doUnpin {
-			unpinMap(m)
+		if doUnpin {
+			mapDef := m.m.def
+			if mapDef.pinning == PIN_CUSTOM_NS {
+				pinPath := options[fmt.Sprintf("maps/%s", m.Name)].PinPath
```
alban: could you check if the entry in the map exists before dereferencing it with `.PinPath`?
```
closeOption, ok := options[fmt.Sprintf("maps/%s", m.Name)]
if !ok {
  return error...
}
pinPath := closeOption.PinPath
```

elf/module.go
dongsupark: @alban Yes, that's what I mean. I'll change the comment to `we don't need to check error for global objects as pinPath would always be an empty string.`

alban: it could still fail for another reason: for example, an user manually unpinned the map by running `/bin/rm` from a terminal. So we can still check the error in the same way as the previous case.

dongsupark: @alban I see your point. To check for the precise error, we would need to distinguish errors of getMapPath from unlink errors. Maybe we would need to define a new error from getMapPath to check for that later, etc. That would complicate the source tree a little more, and would not be exactly the same direction that we took in this PR, right? So I just wrote a precise comment like below:

```
// For the moment we don't need to check error for global objects,
// as pinPath would always be an empty string. Though in the future,
// it would make sense to check for precise error codes to distinguish
// errors of getMapPath() from errors from unlink(2).
```

dongsupark:
@alban @iaguis Updated. Hopefully I addressed every comment.

elf/module.go
dongsupark: @alban Ok, I just squashed this part again into a simple error checking with the help of @schu. Hopefully it would make sense.

elf/module.go
```diff
@@ -504,20 +503,34 @@ func (b *Module) closeSocketFilters() error {
...
+				closeOption, ok := options[fmt.Sprintf("maps/%s", m.Name)]
+				if !ok {
+					return fmt.Errorf("close option for maps/%s must have PinPath set", m.Name)
+				}
+				pinPath = closeOption.PinPath
+			} else if mapDef.pinning == PIN_GLOBAL_NS {
+				// mapDef.namespace is used for PIN_GLOBAL_NS maps
+				pinPath = ""
+			}
+			// TODO: PIN_OBJECT_NS is to be implemented
```
alban: what about 
```
} else if mapDef.pinning == PIN_OBJECT_NS {
    return fmt.Errorf("unpinning with PIN_OBJECT_NS is to be implemented")
}
```

Documentation/pinning.md
```diff
@@ -0,0 +1,81 @@
...
+VFS syscall `unlink(2)`.
+
+Doing that, we can make maps and programs stay alive across process
+terminations. This mechanism provides a much more consistent way of sharing
+objects with other processes, compared to other solutions such as `tc`,
+where objects are shared via Unix domain sockets.
+
+## Different pinning options
+
+`C.bpf_map_def.pinning` can be set to one the following pinning options.
```
alban: Suggestion: add a link:

> `C.bpf_map_def.pinning` (defined in [bpf.h](https://github.com/iovisor/gobpf/blob/master/elf/include/bpf.h#L616)) can be set to one the following pinning options.

bpf_test.go
```diff
@@ -233,6 +237,26 @@ func checkSocketFilters(t *testing.T, b *elf.Module) {
 	}
 }
 
+func checkPinConfig(t *testing.T, b *elf.Module) {
+	var expectedPaths = []string{
+		"/sys/fs/bpf/gobpf-test/testgroup1",
```
alban: same comment as before: parameter `b` is not used there: it could be removed and replaced by `expectedPaths`.

elf/elf.go
```diff
@@ -262,7 +246,16 @@ static int perf_event_open_map(int pid, int cpu, int group_fd, unsigned long fla
 */
 import "C"
 
-const useCurrentKernelVersion = 0xFFFFFFFE
+const (
+	useCurrentKernelVersion = 0xFFFFFFFE
+
+	// Object pin settings should correspond to those of other projects, e.g.:
+	// https://git.kernel.org/pub/scm/linux/kernel/git/shemminger/iproute2.git/tree/include/bpf_elf.h#n25
```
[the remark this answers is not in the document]
alban: (done below :+1: )

Documentation/pinning.md
```diff
@@ -0,0 +1,83 @@
...
+
+Doing that, we can make maps and programs stay alive across process
+terminations. This mechanism provides a much more consistent way of sharing
+objects with other processes, compared to other solutions such as `tc`,
+where objects are shared via Unix domain sockets.
+
+## Different pinning options
+
+`C.bpf_map_def.pinning` (defined in
+[bpf.h](https://github.com/iovisor/gobpf/blob/master/elf/include/bpf.h#L616))
```
iaguis: We shouldn't link to the master branch, line numbers won't match if the code is updated.

elf/elf.go
```diff
@@ -54,26 +54,14 @@ import (
...
-  unsigned int value_size;
-  unsigned int max_entries;
-  unsigned int map_flags;
-  unsigned int pinning;
-  char namespace[BUF_SIZE_MAP_NS];
-} bpf_map_def;
-
 typedef struct bpf_map {
-	int         fd;
+	int fd;
```
iaguis: we don't need this change

Documentation/pinning.md
alban: @iaguis What do you suggest? just remove the link and write `(defined in elf/include/bpf.h)`?

iaguis: Or use https://github.com/iovisor/gobpf/blob/17d86e402/elf/include/bpf.h#L616

dongsupark: @iaguis Yeah, I realized that the line number would not match in the future. But using a specific commit would be also a broken approach, if we force push. I'll try to update it with a specific commit this time, and afterwards I'll try not to force push again in this PR. :-)

iaguis: I think you can merge after this and hopefully nobody force-pushes to master :)

dongsupark: @iaguis Done.

PR #73, opened 2017-08-22
bcc: fix build issues when using bcc 0.4

dongsupark:
(Work-in-progress, please do not merge. After bcc 0.4 has been released, I will get back to this PR.)

Fix build problems in `bcc/symbol.go` and `bcc/perf.go` when using bcc 0.4.
    
* As `bcc_foreach_symbol()` was renamed to `bcc_foreach_function_symbol()` since https://github.com/iovisor/bcc/commit/fcb2ed8e5895, we should conditionally run different functions. bcc04: `bcc_foreach_function_symbol`, !bcc: `bcc_foreach_symbol`

* As prototype of `bcc_resolve_symname()` was changed to include `bcc_symbol_option` since https://github.com/iovisor/bcc/commit/aff6ce7075da, we should contidionally use different function prototypes. bcc04: with `bcc_symbol_option`, !bcc: without `bcc_symbol_option`.

* As prototype of `bcc_open_perf_buffer()` was changed to include `perf_reader_lost_cb` since https://github.com/iovisor/bcc/commit/4b764de6bcb3, we should contidionally use different function prototypes. bcc04: with `perf_reader_lost_cb`, !bcc: without `perf_reader_lost_cb`.

### DIFF
--- a/bcc/perf.go
+++ b/bcc/perf.go
@@ -128,9 +128,8 @@ func InitPerfMap(table *Table, receiverChan chan []byte) (*PerfMap, error) {
 	}
 
 	for _, cpu := range cpus {
-		cpuC := C.int(cpu)
-		reader, err := C.bpf_open_perf_buffer((C.perf_reader_raw_cb)(unsafe.Pointer(C.callback_to_go)), unsafe.Pointer(uintptr(callbackDataIndex)), -1, cpuC, BPF_PERF_READER_PAGE_CNT)
-		if reader == nil {
+		reader, err := bpfOpenPerfBuffer(cpu, callbackDataIndex)
+		if err != nil {
 			return nil, fmt.Errorf("failed to open perf buffer: %v", err)
 		}
 
@@ -179,4 +178,17 @@ func (pm *PerfMap) poll(timeout int) {
 			C.perf_reader_poll(C.int(len(pm.readers)), &pm.readers[0], C.int(timeout))
 		}
 	}
+}
+
+func bpfOpenPerfBuffer(cpu uint, callbackDataIndex uint64) (unsafe.Pointer, error) {
+	cpuC := C.int(cpu)
+	reader, err := C.bpf_open_perf_buffer(
+		(C.perf_reader_raw_cb)(unsafe.Pointer(C.callback_to_go)),
+		nil,
+		unsafe.Pointer(uintptr(callbackDataIndex)),
+		-1, cpuC, BPF_PERF_READER_PAGE_CNT)
+	if reader == nil {
+		return nil, fmt.Errorf("failed to open perf buffer: %v", err)
+	}
+	return reader, nil
 }
--- a/bcc/module.go
+++ b/bcc/module.go
@@ -355,9 +355,9 @@ func (bpf *Module) TableIter() <-chan map[string]interface{} {
 	return ch
 }
 
-func (bpf *Module) attachXDP(devName string, fd int) error {
+func (bpf *Module) attachXDP(devName string, fd int, flags uint32) error {
 	devNameCS := C.CString(devName)
-	res, err := C.bpf_attach_xdp(devNameCS, C.int(fd))
+	res, err := C.bpf_attach_xdp(devNameCS, C.int(fd), C.uint32_t(flags))
 	defer C.free(unsafe.Pointer(devNameCS))
 
 	if res != 0 || err != nil {
@@ -368,10 +368,10 @@ func (bpf *Module) attachXDP(devName string, fd int) error {
 
 // AttachXDP attaches a xdp fd to a device.
 func (bpf *Module) AttachXDP(devName string, fd int) error {
-	return bpf.attachXDP(devName, fd)
+	return bpf.attachXDP(devName, fd, 0)
 }
 
 // RemoveXDP removes any xdp from this device.
 func (bpf *Module) RemoveXDP(devName string) error {
-	return bpf.attachXDP(devName, -1)
+	return bpf.attachXDP(devName, -1, 0)
 }
--- a/bcc/symbol.go
+++ b/bcc/symbol.go
@@ -36,6 +36,12 @@ type symbolAddress struct {
 	addr uint64
 }
 
+type bccSymbolOption struct {
+	useDebugFile      *C.int
+	checkDebugFileCrc *C.int
+	useSymbolType     *C.uint32_t
+}
+
 //symbolCache will cache module lookups
 var symbolCache = struct {
 	cache         map[string][]*symbolAddress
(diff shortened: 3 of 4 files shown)


iaguis:
We mention in [Requirements](https://github.com/iovisor/gobpf#requirements) that we support the latest released version of libbcc. As far as I can see, v0.4.0 was not released yet so I don't think we should merge this.

schu:
An attempt to track bcc#master would be an ongoing maintenance burden while at the same time bring no real value and be error prone. What works today could be broken tomorrow w/o us making any change to the code.

So I also don't think we should do this.

dongsupark:
All right. Then I'll wait until bcc 0.4 has been released.

omac777:
What do I need to do to make this error disappear?  Do I need to resync the gobpf repo?
Do I need to fetch more recent kernel sources?

go build -v -x -work
WORK=/tmp/go-build204912114
github.com/iovisor/gobpf/pkg/cpuonline
mkdir -p $WORK/github.com/iovisor/gobpf/pkg/cpuonline/_obj/
mkdir -p $WORK/github.com/iovisor/gobpf/pkg/
cd /home/dma2/Code/go/src/github.com/iovisor/gobpf/pkg/cpuonline
/home/dma2/go1.9.1/pkg/tool/linux_amd64/compile -o $WORK/github.com/iovisor/gobpf/pkg/cpuonline.a -trimpath $WORK -goversion go1.9.1 -p github.com/iovisor/gobpf/pkg/cpuonline -complete -buildid 478c10d5c546832e818986d91507867fffa1c9bf -D _/home/dma2/Code/go/src/github.com/iovisor/gobpf/pkg/cpuonline -I $WORK -pack ./cpu_range.go
github.com/iovisor/gobpf/bcc
mkdir -p $WORK/github.com/iovisor/gobpf/bcc/_obj/
mkdir -p $WORK/github.com/iovisor/gobpf/
cd /home/dma2/Code/go/src/github.com/iovisor/gobpf/bcc
CGO_LDFLAGS="-g" "-O2" "-lbcc" "-lbcc" "-lbcc" "-lbcc" /home/dma2/go1.9.1/pkg/tool/linux_amd64/cgo -objdir $WORK/github.com/iovisor/gobpf/bcc/_obj/ -importpath github.com/iovisor/gobpf/bcc -- -I $WORK/github.com/iovisor/gobpf/bcc/_obj/ -g -O2 -I/usr/include/bcc/compat -I/usr/include/bcc/compat -I/usr/include/bcc/compat -I/usr/include/bcc/compat module.go perf.go symbol.go table.go
# github.com/iovisor/gobpf/bcc
../../../bcc/symbol.go:94:9: could not determine kind of name for C.bcc_foreach_symbol
root@arch-io-02 2017-10-06_10:56:16_EDT : /home/dma2/Code/go/src/github.com/iovisor/gobpf/examples/bcc/bash_readline

sillyousu:
@omac777 
AFAIK gobpf is lagging behind lastest bcc. 
You can install latest **released version** of bcc.

You can also modify gobpf to work around like me. But it a really quick and dirty hack.
```
diff --git a/bcc/module.go b/bcc/module.go
index d5456bf..b62ece7 100644
--- a/bcc/module.go
+++ b/bcc/module.go
@@ -357,7 +357,7 @@ func (bpf *Module) TableIter() <-chan map[string]interface{} {

 func (bpf *Module) attachXDP(devName string, fd int) error {
        devNameCS := C.CString(devName)
-       res, err := C.bpf_attach_xdp(devNameCS, C.int(fd))
+       res, err := C.bpf_attach_xdp(devNameCS, C.int(fd), 0)
        defer C.free(unsafe.Pointer(devNameCS))

        if res != 0 || err != nil {
diff --git a/bcc/perf.go b/bcc/perf.go
index a011a50..9d6d72e 100644
--- a/bcc/perf.go
+++ b/bcc/perf.go
@@ -129,7 +129,7 @@ func InitPerfMap(table *Table, receiverChan chan []byte) (*PerfMap, error) {

        for _, cpu := range cpus {
                cpuC := C.int(cpu)
-               reader, err := C.bpf_open_perf_buffer((C.perf_reader_raw_cb)(unsafe.Pointer(C.callback_to_go)), unsafe.Pointer(uintptr(callbackDataIndex)), -1, cpuC, BPF_PERF_READER_PAGE_CNT)
+               reader, err := C.bpf_open_perf_buffer((C.perf_reader_raw_cb)(unsafe.Pointer(C.callback_to_go)), nil, unsafe.Pointer(uintptr(callbackDataIndex)), -1, cpuC, BPF_PERF_READER_PAGE_CNT)
                if reader == nil {
                        return nil, fmt.Errorf("failed to open perf buffer: %v", err)
                }
diff --git a/bcc/symbol.go b/bcc/symbol.go
index 7d6ffdc..0cbb52d 100644
--- a/bcc/symbol.go
+++ b/bcc/symbol.go
@@ -67,7 +67,7 @@ func resolveSymbolPath(module string, symname string, addr uint64, pid int) (str
                pid = 0
        }

-       res, err := C.bcc_resolve_symname(moduleCS, symnameCS, (C.uint64_t)(addr), C.int(pid), symbolC)
+       res, err := C.bcc_resolve_symname(moduleCS, symnameCS, (C.uint64_t)(addr), C.int(pid), nil, symbolC)
        if res == 0 {
                return C.GoString(symbolC.module), (uint64)(symbolC.offset), nil
        }
@@ -91,7 +91,7 @@ func getUserSymbolsAndAddresses(module string) ([]*symbolAddress, error) {

        moduleCS := C.CString(module)
        defer C.free(unsafe.Pointer(moduleCS))
-       res := C.bcc_foreach_symbol(moduleCS, (C.SYM_CB)(unsafe.Pointer(C.foreach_symbol_callback)))
+       res := C.bcc_foreach_function_symbol(moduleCS, (C.SYM_CB)(unsafe.Pointer(C.foreach_symbol_callback)))
        if res < 0 {
                return nil, fmt.Errorf("unable to list symbols for %s", module)
        }
```

schu:
@omac777 @sillyousu correct, gobpf master tracks the [latest release of bcc](https://github.com/iovisor/bcc/releases), i.e. `v0.3.0` currently. `v0.3.0` is from March; if you want you could ask the bcc maintainers to tag a new release which we could update to then.

schu:
[v0.4.0](https://github.com/iovisor/bcc/releases/tag/v0.4.0) is out now.

@dongsupark do you have time and interest to pick up this PR? I think most changes are not needed, since we don't want to support multiple bcc versions, i.e. it's only about updating places where the bcc API has changed.

dongsupark:
@schu Of course. I'm going to look into this PR next week.
Thanks for letting me know!

dongsupark:
Updated the PR.
* Added a commit to fix another build issue in bcc/module.go
* Simply removed the old code based on bcc 0.3, and updated the code to be based on bcc 0.4.

Still the CI fails. Its reason: Previously `docker://schu/gobpf-ci` was based on Fedora 25 (bcc 0.3), so the new code won't work with the docker image. We could update the gobpf-ci image to have an additional bcc 0.4 package. Though `repo.iovisor.org` still does not provide bcc 0.4. So the bcc needs to be somehow manually built to be included in the gobpf-ci image.

dongsupark:
Created a PR to update the gobpf-ci image: https://github.com/schu/docker-gobpf-ci/pull/1

bcc/perf.go
```diff
@@ -180,3 +180,25 @@ func (pm *PerfMap) poll(timeout int) {
 		}
 	}
 }
+
+//export callback_to_lost
+func callback_to_lost(lost C.uint64_t) {
```
schu: Does this work? IIRC `callback_to_lost` gets passed the number of lost events, not a cookie value that could be used to lookup `callbackData`. Also, lost events should be send over a separate channel (in the elf package we accept a `lostChan`).

dongsupark: @schu Yeah, you're right. I made it use `lostChan`, and updated relevant examples too. Force pushed.

schu: Adding a `lostChan` alone unfortunately isn't sufficient and `callbackData := lookupCallback(uint64(lost))` won't work, as we don't have the [`callbackDataIndex`](https://github.com/kinvolk/gobpf/blob/161be1f8caaddc56bd6577187f867c5057c327f9/bcc/perf.go#L117) in `callback_to_lost`, i.e. we need to find a different way to expose this to the user.

I would be ok with setting the callback to `nil` in this pull request and open a follow-up issues to address that separately.

examples/bcc/bash_readline/bash_readline.go
```diff
@@ -82,19 +83,20 @@ func main() {
...
 
-	fmt.Printf("%10s\t%s\n", "PID", "COMMAND")
+	fmt.Printf("%10s\t%s\t%d\n", "PID", "COMMAND", "LOST")
 	go func() {
 		var event readlineEvent
 		for {
-			data := <-channel
-			err := binary.Read(bytes.NewBuffer(data), binary.LittleEndian, &event)
+			receiverData := <-receiverChan
+			lostData := <-lostChan
```
schu: This doesn't work as intended: the example does block here until events are lost. Should be a `select` instead.

But same thing as above: could be addressed in a separate pull request.

dongsupark: @schu Done. I simply set lost_cb to `nil`. Now it should work.

bcc/perf.go
```diff
@@ -180,3 +180,16 @@ func (pm *PerfMap) poll(timeout int) {
 		}
 	}
 }
+
+func bpfOpenPerfBuffer(cpu uint, callbackDataIndex uint64) (unsafe.Pointer, error) {
+	cpuC := C.int(cpu)
+	reader, err := C.bpf_open_perf_buffer(
+		(C.perf_reader_raw_cb)(unsafe.Pointer(C.callback_to_go)),
+		(C.perf_reader_lost_cb)(unsafe.Pointer(nil)),
```
schu: Isn't just `nil` enough?

bcc/perf.go
```diff
@@ -33,6 +33,7 @@ import (
 // perf_reader_raw_cb as defined in bcc libbpf.h
 // typedef void (*perf_reader_raw_cb)(void *cb_cookie, void *raw, int raw_size);
 extern void callback_to_go(void*, void*, int);
+// typedef void (*perf_reader_lost_cb)(uint64_t lost);
```
schu: Can this be removed as well?

bcc/perf.go
dongsupark: @schu Done. I changed it simply to `nil` without perf_reader_lost_cb.

PR #76, opened 2017-08-28
Add support for SCHED_{CLS,ACT} progs

schu:


alban:
It should have examples explaining how to attach it to a qdisc. Since the PR does not have functions to hook the bpf programs to a qdisc, I guess you can do it manually by pinning the programs on bpffs and calling `tc`? Did they add parameters for that in the `tc` CLI?

schu:
> I guess you can do it manually by pinning the programs on bpffs and calling tc? Did they add parameters for that in the tc CLI?

`object-pinned` is supported. I will ll try to put together a small example.

schu:
>> I guess you can do it manually by pinning the programs on bpffs and calling tc? Did they add parameters for that in the tc CLI?

> object-pinned is supported. I will ll try to put together a small example.

I have updated the commit message (d16f63bd8810465ee11c8f8c46131b5215e1d106) with a tc example.

I have also added 4f9e9bc784957dcbd68d1d4859c376acff00b445 as an alternative way to pin an object.

PR #78, opened 2017-08-28
elf: increase size of log buffer again

iaguis:
0e0349c already increased the log buffer but while testing on kernel
4.4 I ran out of buffer space again.

Let's try with 512k. Hopefully this is enough.


schu:
LGTM. But I wonder if we should remove `log` from `Module` and instead have a `LoadWithLog` method or similar for users who wish to have a log.

iaguis:
> LGTM. But I wonder if we should remove log from Module and instead have a LoadWithLog method or similar for users who wish to have a log.

I filed https://github.com/iovisor/gobpf/issues/80

PR #79, opened 2017-08-29
elf,bpf_test: Improve error checks

dongsupark:
Remove an unnecessary error check in `elfReadVersion()`. The error is already being checked above in the same function, https://github.com/iovisor/gobpf/blob/master/elf/elf.go#L284

We should check for a specific error code `os.ErrInvalid`, instead of comparing error strings with each other.

### DIFF
--- a/bpf_test.go
+++ b/bpf_test.go
@@ -17,7 +17,6 @@
 package bpf
 
 import (
-	"strings"
 	"testing"
 
 	"github.com/iovisor/gobpf/bcc"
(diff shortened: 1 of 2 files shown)


bpf_test.go
```diff
@@ -255,7 +255,7 @@ func TestModuleLoadELF(t *testing.T) {
 		t.Fatal("prog is nil")
 	}
 	if err := b.Load(nil); err != nil {
-		if !strings.Contains(err.Error(), "invalid argument") {
```
schu: I wonder why we ignore the error at all? @iaguis do you remember? Changed with 8f18c857026f0495eed5b00c28f9697548d88519

iaguis: I don't really know :/

I think we can just remove it.

dongsupark: @iaguis Removed the check for invalid argument.

schu:
Thanks!

PR #81, opened 2017-08-30
Added XDP support

GustavoKatel:
Hi. Congrats for the great work so far.

This PR adds XDP support based on the python-bcc lib.

Please let me know if there's any change to be made.

### DIFF
--- a/bcc/module.go
+++ b/bcc/module.go
@@ -351,3 +351,24 @@ func (bpf *Module) TableIter() <-chan map[string]interface{} {
 	}()
 	return ch
 }
+
+func (bpf *Module) attachXDP(devName string, fd int) error {
+	devNameCS := C.CString(devName)
+	res, err := C.bpf_attach_xdp(devNameCS, C.int(fd))
+	defer C.free(unsafe.Pointer(devNameCS))
+
+	if res != 0 || err != nil {
+		return fmt.Errorf("failed to attach BPF xdp to device %v: %v", devName, err)
+	}
+	return nil
+}
+
+// AttachXDP attaches a xdp fd to a device.
+func (bpf *Module) AttachXDP(devName string, fd int) error {
+	return bpf.attachXDP(devName, fd)
+}
+
+// RemoveXDP removes any xdp from this device.
+func (bpf *Module) RemoveXDP(devName string) error {
+	return bpf.attachXDP(devName, -1)
+}
--- a/examples/bcc/xdp/xdp_drop.go
+++ b/examples/bcc/xdp/xdp_drop.go
@@ -1,0 +1,178 @@
+// xdp_drop.go Drop incoming packets on XDP layer and count for which
+// protocol type. Based on:
+// https://github.com/iovisor/bcc/blob/master/examples/networking/xdp/xdp_drop_count.py
+//
+// Copyright (c) 2017 GustavoKatel
+// Licensed under the Apache License, Version 2.0 (the "License")
+
+package main
+
+import (
+	"fmt"
+	"os"
+	"os/signal"
+	"strconv"
+
+	bpf "github.com/iovisor/gobpf/bcc"
+)
+
+/*
+#cgo CFLAGS: -I/usr/include/bcc/compat
+#cgo LDFLAGS: -lbcc
+#include <bcc/bpf_common.h>
+#include <bcc/libbpf.h>
+void perf_reader_free(void *ptr);
+*/
+import "C"
+
+const source string = `
+#define KBUILD_MODNAME "foo"
+#include <uapi/linux/bpf.h>
+#include <linux/in.h>
+#include <linux/if_ether.h>
+#include <linux/if_packet.h>
+#include <linux/if_vlan.h>
+#include <linux/ip.h>
+#include <linux/ipv6.h>
+
+BPF_TABLE("array", int, long, dropcnt, 256);
+
+static inline int parse_ipv4(void *data, u64 nh_off, void *data_end) {
+    struct iphdr *iph = data + nh_off;
+
+    if ((void*)&iph[1] > data_end)
+        return 0;
+    return iph->protocol;
+}
+
+static inline int parse_ipv6(void *data, u64 nh_off, void *data_end) {
+    struct ipv6hdr *ip6h = data + nh_off;
+
+    if ((void*)&ip6h[1] > data_end)
+        return 0;
+    return ip6h->nexthdr;
+}
+
+int xdp_prog1(struct CTXTYPE *ctx) {
+
+    void* data_end = (void*)(long)ctx->data_end;
+    void* data = (void*)(long)ctx->data;
+
+    struct ethhdr *eth = data;
+
+    // drop packets
+    int rc = RETURNCODE; // let pass XDP_PASS or redirect to tx via XDP_TX
+    long *value;
+    uint16_t h_proto;
+    uint64_t nh_off = 0;
+    int index;
+
+    nh_off = sizeof(*eth);
+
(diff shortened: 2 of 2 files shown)


bcc/module.go
```diff
@@ -351,3 +351,24 @@ func (bpf *Module) TableIter() <-chan map[string]interface{} {
...
 	return ch
 }
+
+func (bpf *Module) attachXDP(devName string, fd int) error {
+	devNameCS := C.CString(devName)
+	res, err := C.bpf_attach_xdp(devNameCS, C.int(fd))
+	C.free(unsafe.Pointer(devNameCS))
+
+	if res != 0 || err != nil {
+		return fmt.Errorf("failed to attach BPF xdp to device %v: %v", err, devName)
```
nhlfr: Please change the order of variables. You probably want to format `devName` first, then `err`.

bcc/module.go
```diff
@@ -351,3 +351,24 @@ func (bpf *Module) TableIter() <-chan map[string]interface{} {
 	}()
 	return ch
 }
+
+func (bpf *Module) attachXDP(devName string, fd int) error {
+	devNameCS := C.CString(devName)
+	res, err := C.bpf_attach_xdp(devNameCS, C.int(fd))
+	C.free(unsafe.Pointer(devNameCS))
```
nhlfr: I'd prefer to `defer` all `C.free` operations - it's more readable.

examples/bcc/xdp/xdp_drop.go
```diff
@@ -0,0 +1,179 @@
...
+		if value > 0 {
+			fmt.Printf("%v: %v pkts\n", key, value)
+		}
+	}
+
+	err = module.RemoveXDP(device)
+	if err != nil {
+		fmt.Println(err)
+	}
+
```
nhlfr: Can we remove this empty line?

examples/bcc/xdp/xdp_drop.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+
+		if value > 0 {
+			fmt.Printf("%v: %v pkts\n", key, value)
+		}
+	}
+
+	err = module.RemoveXDP(device)
+	if err != nil {
+		fmt.Println(err)
```
nhlfr: I'd prefer to print errors to stderr instead of stdout, so

```
fmt.Fprintf(os.Stderr, err)
```

examples/bcc/xdp/xdp_drop.go
```diff
@@ -0,0 +1,179 @@
...
+	fn, err := module.Load("xdp_prog1", mode)
+
+	if mode == C.BPF_PROG_TYPE_XDP {
+		err = module.AttachXDP(device, fn)
+
+		if err != nil {
+			fmt.Println(err)
+			os.Exit(1)
+		}
+
```
nhlfr: Can we remove this empty line?

examples/bcc/xdp/xdp_drop.go
```diff
@@ -0,0 +1,179 @@
...
+`
+
+func usage() {
+	fmt.Printf("Usage: %v <ifdev>\n", os.Args[0])
+	fmt.Printf("e.g.: %v eth0\n", os.Args[0])
+	os.Exit(1)
+}
+
+func main() {
+
```
nhlfr: Can we remove this empty line?

GustavoKatel:
Thanks for the comments and suggestions @nhlfr Everything looks good now?

schu:
The example `xdp_drop.go` doesn't give any output, that part is not working for me (probably due to `Table.Iter()` using `..snprintf` and returning hex digits as string: https://github.com/iovisor/gobpf/blob/master/bcc/table.go#L204-L215).

@GustavoKatel what data do you get there that the `strconv.ParseUint` works for you?

Please credit the authors/source of the example program in a comment: https://github.com/iovisor/bcc/blob/master/examples/networking/xdp/xdp_drop_count.py

Also, can you please squash fixup commits? In this particular case, it should be only two commits, something like:

```
bcc: add xdp support
examples: add bcc xdp_drop example
```

examples/bcc/xdp/xdp_drop.go
```diff
@@ -0,0 +1,176 @@
...
+func main() {
+	var device string
+
+	if len(os.Args) != 2 {
+		usage()
+	}
+
+	device = os.Args[1]
+
+	mode := C.BPF_PROG_TYPE_XDP
```
schu: Can you remove `mode` and instead use `C.BPF_PROG_TYPE_XDP` directly in `Load`. No need for the `if mode == C.BPF_PROG_TYPE_XDP {` check below then.

examples/bcc/xdp/xdp_drop.go
```diff
@@ -0,0 +1,176 @@
...
+		if err != nil {
+			continue
+		}
+
+		if value > 0 {
+			fmt.Printf("%v: %v pkts\n", key, value)
+		}
+	}
+
+	err = module.RemoveXDP(device)
```
schu: Instead, you could a `defer` after attaching:

```
defer func() {
    if err := module.RemoveXDP(device); err != nil {
        fmt.Fprintf(os.Stderr, "Failed to remove XDP from %s: %v\n", device, err)
    }
}()
```

GustavoKatel:
Changes as requested, and squashed commits.

@schu Can you elaborate how did you test it?

I'm using https://github.com/iovisor/xdp-vagrant to emulate a supported driver.

Also the `dropcnt` table contains an `uint32_t` index and values are `long`.

Please let me know if there's anything else I can do. Thanks.

schu:
Thank you for the update :)

> Can you elaborate how did you test it?

```
$ sudo -E go run examples/bcc/xdp/xdp_drop.go wlp4s0 
Dropping packets, hit CTRL+C to stop
^C
{IP protocol-number}: {total dropped pkts}
$
```

ie. packets are dropped, but I don't get the expected output. The bcc example (`./examples/networking/xdp/xdp_drop_count.py wlp4s0`) works.

By adding the following patch, I can see in `trace_pipe` that values are counted correctly on kernel side.

```
diff --git a/examples/bcc/xdp/xdp_drop.go b/examples/bcc/xdp/xdp_drop.go
index 9ef4c35..e905854 100644
--- a/examples/bcc/xdp/xdp_drop.go
+++ b/examples/bcc/xdp/xdp_drop.go
@@ -102,8 +102,10 @@ int xdp_prog1(struct CTXTYPE *ctx) {
         index = 0;
 
     value = dropcnt.lookup(&index);
-    if (value)
+    if (value) {
         *value += 1;
+        bpf_trace_printk("index: %d value: %ld\n", index, *value);
+    }
 
     return rc;
 }
```

I'm trying to find out what the problem is.

examples/bcc/xdp/xdp_drop.go
```diff
@@ -0,0 +1,180 @@
...
+	for entry := range dropcnt.Iter() {
+		var key, value uint64
+		var err error
+
+		key, err = strconv.ParseUint(entry.Key, 0, 32)
+		if err != nil {
+			continue
+		}
+
+		value, err = strconv.ParseUint(entry.Value, 0, 64)
```
alban: Is it the correct way to get the value for each CPU?
This page suggests the lookup returns an array with the values for each CPU:

https://www.mail-archive.com/netdev@vger.kernel.org/msg95249.html

GustavoKatel:
Hi guys, 

I guess @alban was in the right direction with the percpu array.

I've changed to a normal array and now the sync fetch and add is done in the xdp prog instead.

I hope it's ok now. Please let me know.

Thanks!

PR #83, opened 2017-09-07
examples: avoid padding null bytes in command string

tklauser:
Avoid having padding null chars by using u32/uint32 consistently for the
PID and by packing the struct.

Also remove the unnecessary import "C", the example itself is not using
any cgo.

Signed-off-by: Tobias Klauser <tklauser@distanz.ch>


schu:
Thanks @tklauser!

PR #91, opened 2017-10-16
bcc: add Table.DeleteAll

Pryz:
This PR adds a `DeleteAll` function to be able to easily clear all keys of a table. Idea is to have something like https://github.com/iovisor/bcc/blob/master/docs/reference_guide.md#5-clear.

Not quite sure what's the contribution process so let me know if there is something else I have to do here.

### DIFF
--- a/bcc/table.go
+++ b/bcc/table.go
@@ -164,6 +164,23 @@ func (table *Table) Delete(keyStr string) error {
 	return nil
 }
 
+// Delete all keys from the table
+func (table *Table) DeleteAll() error {
+	mod := table.module.p
+	fd := C.bpf_table_fd_id(mod, table.id)
+
+	keySize := C.bpf_table_key_size_id(mod, table.id)
+	key := make([]byte, keySize)
+	keyP := unsafe.Pointer(&key[0])
+	for res := C.bpf_get_first_key(fd, keyP, keySize); res == 0; res = C.bpf_get_next_key(fd, keyP, keyP) {
+		r, err := C.bpf_delete_elem(fd, keyP)
+		if r != 0 {
+			return fmt.Errorf("Table.DeleteAll: unable to delete element: %v", err)
+		}
+	}
+	return nil
+}
+
 // Iter returns a receiver channel to iterate over all table entries.
 func (table *Table) Iter() <-chan Entry {
 	mod := table.module.p
--- a/bpf_test.go
+++ b/bpf_test.go
@@ -70,6 +70,38 @@ func TestModuleLoadBCC(t *testing.T) {
 	}
 }
 
+func TestBCCTableDeleteAll(t *testing.T) {
+	b := bcc.NewModule(simple1, []string{})
+	if b == nil {
+		t.Fatal("prog is nil")
+	}
+	defer b.Close()
+	table := bcc.NewTable(b.TableId("table1"), b)
+	if err := table.Set("1", "11"); err != nil {
+		t.Fatalf("table.Set failed: %v", err)
+	}
+	if err := table.Set("2", "22"); err != nil {
+		t.Fatalf("table.Set failed: %v", err)
+	}
+	count := 0
+	for _ = range table.Iter() {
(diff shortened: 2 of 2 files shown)


bcc/table.go
```diff
@@ -164,6 +164,24 @@ func (table *Table) Delete(keyStr string) error {
 	return nil
 }
 
+// Delete all keys from the table
+func (table *Table) DeleteAll() error {
+	mod := table.module.p
+	fd := C.bpf_table_fd_id(table.module.p, table.id)
```
schu: You could also use `mod` here for `table.module.p`

bcc/table.go
```diff
@@ -164,6 +164,24 @@ func (table *Table) Delete(keyStr string) error {
...
+
+	key_size := C.bpf_table_key_size_id(mod, table.id)
+	key := make([]byte, key_size)
+	keyP := unsafe.Pointer(&key[0])
+	res := C.bpf_lookup_elem(fd, keyP, nil)
+	for res = C.bpf_get_next_key(fd, keyP, keyP); res == 0; res = C.bpf_get_next_key(fd, keyP, keyP) {
```
schu: bcc now has a function [`bpf_get_first_key`](https://github.com/iovisor/bcc/blob/fdf9b08cee7844d9ad86cbb1ab6b5adeff2b0520/src/cc/libbpf.c#L141) which takes care of handling kernel specific peculiarities. I would prefer we wait for the [update to bcc v0.4.0](https://github.com/iovisor/gobpf/pull/73). After that, you could do the `for` as follows:

```
for res = C.bpf_get_first_key(fd, keyP, keySize); res == 0; res = C.bpf_get_next_key(fd, keyP, keyP) {
```

bcc/table.go
```diff
@@ -164,6 +164,24 @@ func (table *Table) Delete(keyStr string) error {
...
+		r, err := C.bpf_delete_elem(fd, keyP)
+		if r != 0 {
+			return fmt.Errorf("Table.DeleteAll: unable to delete element : %v", err)
```
schu: Stray ` ` before the colon.

schu:
> Not quite sure what's the contribution process so let me know if there is something else I have to do here.

Please prefix your commit message title with `bcc: ...`, otherwise all good :) Thanks!

schu:
gobpf is now updated to bcc v0.4.0.

Pryz:
rebased and updated :)

bcc/table.go
```diff
@@ -164,6 +164,24 @@ func (table *Table) Delete(keyStr string) error {
...
 
+// Delete all keys from the table
+func (table *Table) DeleteAll() error {
+	mod := table.module.p
+	fd := C.bpf_table_fd_id(mod, table.id)
+
+	keySize := C.bpf_table_key_size_id(mod, table.id)
+	key := make([]byte, keySize)
+	keyP := unsafe.Pointer(&key[0])
+	res := C.bpf_lookup_elem(fd, keyP, nil)
```
schu: Is this `bpf_lookup_elem` still needed?

nathanleclaire:
Why not call it `Clear` for consistency with existing Python bindings?

Pryz:
Any specific reason I guess I could rename it :)

@schu thought ?

Pryz:
@schu I would need some advice for the test. I've pushed a test with a specific bpf program but I'm pretty sure there is an easier way to test that.

schu:
> I would need some advice for the test. I've pushed a test with a specific bpf program but I'm pretty sure there is an easier way to test that.

You could re-use the existing `table1` like follows:

https://github.com/kinvolk/gobpf/commit/544211703374661ef700c4dafba6e2811ea3ba37

I'm fine with both `DeleteAll` and `Clear`, as you prefer.

Pryz:
Oh nice. I've totally missed the `Set` function. Test updated !

Pryz:
Thanks !

PR #100, opened 2017-12-21
Update to bcc 0.5.0

schu:
Note that `bpf_prog_load` as in bcc 0.5.0 prints the log buf to stderr in any case, even when successful, i.e. there's no way currently to suppress output as far as I have seen. In master this was already changed to only print when `ret < 0`: https://github.com/iovisor/bcc/commit/c669257c1f16eae4c695866bea147f57a9092b80 Ideally, we would be able to disable output entirely.

### DIFF
--- a/bcc/module.go
+++ b/bcc/module.go
@@ -47,6 +47,8 @@ type compileRequest struct {
 	rspCh  chan *Module
 }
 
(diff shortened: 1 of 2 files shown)


schu:
Thanks @dongsupark

Issue #89, opened 2017-10-16
Table Entry vs Get/Set/Delete


Pryz:
Hi there,

Based on the doc, I was expected to be able to do things like : 
```
for e := range table.Iter() {
  ...
  table.Delete(e.Key)
}
```

But I am getting this error : 
```
Table.Delete: unable to delete element ({ 0xffffffff88dbecd1 }): no such file or directory
```
I'm not sure what I'm missing here. I'm using a `BPF_HASH`.
What's the difference between `keyStr` from [Get](https://godoc.org/github.com/iovisor/gobpf/bcc#Table.Get) or Delete/Set and `Key` of an [Entry](https://godoc.org/github.com/iovisor/gobpf/bcc#Entry) ?

Thanks !

schu:
Hey, `table.Iter()` uses [bpf_table_key_snprintf](https://github.com/iovisor/gobpf/blob/master/bcc/table.go#L204) for the key which returns a string representation of the key according to the key type of the table (https://github.com/iovisor/bcc/blob/91837cacfbe74141d88efcd096ff8d33a1517d2d/src/cc/bpf_module.cc#L515-L516)

The `table.Iter()`/`table.Entry` usage has led to confusion in the past. I would also like to see this changed to something more straightforward.

schu:
So for the time being you somehow have to find your way around that to delete all entries.

Pryz:
Make sense. Any thought about https://github.com/iovisor/gobpf/pull/91 ?

PR #102, opened 2017-12-28
Move to retrieving byte slices from BCC table entries instead of formatted strings

nathanleclaire:
This PR updates `table.Entry` to contain a type of `[]byte` instead of `string`. This is due to confusion (e.g., https://github.com/iovisor/gobpf/issues/89) around the fact that these `string` were obtained from the `bpf_table_key_snprintf` function. `[]byte` does not require any hex decoding etc. like the `string`, just figuring out how to get the returned binary information into the proper Go type.

Also, things like `Clear` / `DeleteAll` (https://github.com/iovisor/gobpf/pull/91) become way more straightforward, because you have the verbatim bytes of the key in the first place, so it's easy to go back and delete a specific leaf.

It also adds methods `bccencoding.Unmarshal`, `entry.UnmarshalKey`, and `entry.UnmarshalValue` for fast iteration and prototyping of getting information from BCC. They do use reflection, but faster methods will always be available to users if desired because they can `binary.Read`, etc. Eventually code generators for this, such as an equivalent to https://github.com/pquerna/ffjson as compared to `encoding/json`, could be added as well if desired.

I do want to add another example and tests demonstrating struct unmarshalling, etc. `bccencoding.UnmarshalBytes` just covers `uint32`, `uint64`, `string`, and `struct` composed of those today.

cc @schu @iaguis  and also FYI @Pryz you might find this interesting 

**Notes:**

- [x] ~Is `bccencoding` a good package name with path of `github.com/iovisor/gobpf/bcc/encoding`? This doesn't seem to play nice with `goimports` so I might change the directory name.~
- [x] ~Maybe `bccencoding.UnmarshalBytes` should just be `Unmarshal`?~
- [ ] Unit tests still needed
- [ ] Adding more types (nested struct support, etc.)

### DIFF
--- a/bcc/table.go
+++ b/bcc/table.go
@@ -15,9 +15,10 @@
 package bcc
 
 import (
-	"bytes"
 	"fmt"
 	"unsafe"
+
+	"github.com/iovisor/gobpf/bcc/bccencoding"
 )
 
 /*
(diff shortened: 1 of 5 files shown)


bcc/table.go
```diff
@@ -180,40 +144,40 @@ func (table *Table) Iter() <-chan Entry {
...
-			res = C.bpf_lookup_elem(fd, keyP, leafP)
 		}
 		if res == 0 {
 			return
 		}
-		keyStr := make([]byte, key_size*8)
-		leafStr := make([]byte, leaf_size*8)
-		keyStrP := (*C.char)(unsafe.Pointer(&keyStr[0]))
-		leafStrP := (*C.char)(unsafe.Pointer(&leafStr[0]))
 		for res = C.bpf_get_next_key(fd, keyP, keyP); res == 0; res = C.bpf_get_next_key(fd, keyP, keyP) {
```
schu: bcc has a function [`bpf_get_first_key`](https://github.com/iovisor/bcc/blob/fdf9b08cee7844d9ad86cbb1ab6b5adeff2b0520/src/cc/libbpf.c#L141) now, so you could also remove the initialization with an invalid key above and reduce this to

```
for res = C.bpf_get_first_key(fd, keyP, keySize); res == 0; res = C.bpf_get_next_key(fd, keyP, keyP) {
```

gimpf:
Hi @nathanleclaire! Are you really sure that it is best to include packing/unpacking plain structs to byte slices as part of gobpf? This is a generally useful feature, and there seems to be another library that should do that just fine (see [lunixbochs/struc](https://github.com/lunixbochs/struc/))

nathanleclaire:
Hi @gimpf !

> Are you really sure that it is best to include packing/unpacking plain structs to byte slices as part of gobpf? 

I don't have any real preference - I defer to what the maintainers think is best. @schu any opinions?

Given how hairy it can be, I'm definitely in favor of using a 3rd party dep if it's relatively sane and minimal. It might even fix a few bugs that I suspect exist with this patch.

dswarbrick:
At the moment, I have to resort to shenanigans like this, to get items into a usable form after gobpf has "helpfully" `snprintf`'ed them to bytes, and _then_ returned that as a string.
```go
// parseKey parses a BPF hash key as created by the BPF program. For example:
// `{ "sda" 0xb }` is the key for the 11th bucket for device "sda".
func parseKey(s string) (string, uint64) {
	fields := strings.Fields(strings.Trim(s, "{ }"))
	label := strings.Trim(fields[0], "\"")
	bucket, _ := strconv.ParseUint(fields[1], 0, 64)
	return label, bucket
}
```
That was for a BPF program with the following key struct:
```c
typedef struct disk_key {
	char disk[DISK_NAME_LEN];
	u64 slot;
} disk_key_t;
```
It would be great to simply be able to access to those struct members at a lower level, either as a byte slice, or `[]interface{}` of struct members. It seems silly for gobpf to `snprintf` stuff, only for me to effectively `sscanf` in Go. The same goes for the value - I have to `strconv.ParseUint(entry.Value, 0, 64)`. Why can't I simply get access to the key / leaf in their original form?

bobrik:
Any updates here?

gimpf:
@bobrik Some: I've updated PR #112 (better table iterator) to integrate parts of this PR (`[]byte` based Get/Set). I skipped the reflection based unpacking into structs, as I think a separate library (as linked above, for instance [lunixbochs/struc](https://github.com/lunixbochs/struc/) is a better approach.

I hope that PR #112 will make it into master, and intend to fix any code-review issues from the maintainers should such come up. If you have a need for/prefer []byte based access, you might also want to have a look into that one; should you find something amiss, we should fix that before, as it does break the API.

schu:
I prefer going with #112 when ready and leave packing/unpacking out of gobpf and to the user.

> It seems silly for gobpf to snprintf stuff, only for me to effectively sscanf in Go.

This code goes back to [iomodules](https://github.com/iovisor/iomodules/tree/master/hover/bpf) (which gobpf was partly based on) and was never updated for use in a library really. So the work by @nathanleclaire and @gimpf to change that is very welcome :)

nathanleclaire:
I like #112 - would be happy to see that one land, close this PR, and use a 3rd party for unpacking like mentioned.

schu:
> I like #112 - would be happy to see that one land, close this PR, and use a 3rd party for unpacking like mentioned.

Thanks @nathanleclaire - #112 was merged.

PR #104, opened 2018-01-02
AttachUretprobe prefix with r_, not p_

NathanFrench:
AttachUretprobe must prefix with `r_`, not `p_` in order for uretprobes to work.

With this stupid patch, the probes are now correct (via perf probe -l)
```
$ sudo perf probe -l
  uprobes:p__usr_local_lib_XXX_bcc_XXX (on SomeFunction in /usr/local/lib/libXXX)
  uprobes:r__usr_local_lib_XXX_bcc_XXX (on SomeFunction%return in /usr/local/lib/libXXX)
```

### DIFF
--- a/bcc/module.go
+++ b/bcc/module.go
@@ -289,7 +289,7 @@ func (bpf *Module) AttachUretprobe(name, symbol string, fd, pid int) error {
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


schu:
Thanks @NathanFrench

PR #105, opened 2018-01-03
pkg: Add Ksym helper

Pryz:
This PR adds a helper to retrieve the name of a kernel function based on its address.

Context : to implement some monitoring tools I needed something like https://github.com/iovisor/bcc/blob/master/docs/reference_guide.md#1-ksym.

Not sure if reading `/proc/kallsyms` is the best way to do it. Please let me know.

### DIFF
--- a/pkg/ksym/ksym.go
+++ b/pkg/ksym/ksym.go
@@ -1,0 +1,48 @@
+package ksym
+
+import (
+	"bufio"
+	"errors"
+	"io"
+	"os"
+	"strings"
+)
+
+const (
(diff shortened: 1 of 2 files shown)


schu:
Since the code is not used in gobpf, I feel it better lives as a package in your tool.

Pryz:
Sounds good. I thought it would be useful to have some of the same helpers coverage that the python library offers.

Closing then :)

Pryz:
@schu What do you think about creating a "helper" package in gobpf ?

PR #106, opened 2018-01-05
bcc: make logging configurable

### LINKED ISSUE #103: Module.load() failing with ENOSPC due to too small log buffer
When trying to load a relatively simple BPF program (one hash, four histograms), Module.load() fails due to a statically configured and too small log buffer:
```go
    logbuf := make([]byte, 65536)
    logbufP := (*C.char)(unsafe.Pointer(&logbuf[0]))
    fd, err := C.bpf_prog_load(uint32(progType), nameCS, start, size, license, version, defaultLogLevel, logbufP, C.uint(len(logbuf)))
```
Doubling the size of the `logbuf` slice or monkey-patching `defaultLogLevel` to be zero allows the program to load.

I haven't seen this issue when trying to load the same BPF program using the Python BCC libs. Either way, it would be great to be able to specify a larger log buffer, or turn off logging, without having to monkey-patch gopbf.

schu:
Currently, logging is always enabled (log level `1`) and the log buffer
set to a fixed size of 65536 bytes. Since the log buffer must be large
enough for all verifier message when logging is enabled, the code
returns `ENOSPC` for larger programs and doesn't allow the user to
disable logging or increase the buffer.

Instead, disable bpf logging by default and allow the user to specify
log level and buffer size on `Load()`. This way, users can specify a
large enough buffer matching their program if desired.

Also, add a missing error check in `examples/bcc/xdp/xdp_drop.go`.

Resolves #103

### DIFF
--- a/bcc/module.go
+++ b/bcc/module.go
@@ -15,7 +15,6 @@
 package bcc
 
 import (
-	"bytes"
 	"fmt"
 	"regexp"
 	"runtime"
@@ -46,8 +45,6 @@ type compileRequest struct {
 	cflags []string
 	rspCh  chan *Module
 }
-
-const defaultLogLevel = 1
 
 const (
 	BPF_PROBE_ENTRY = iota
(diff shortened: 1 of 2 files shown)


bcc/module.go
```diff
@@ -135,34 +132,34 @@ func (bpf *Module) Close() {
 
 // LoadNet loads a program of type BPF_PROG_TYPE_SCHED_ACT.
 func (bpf *Module) LoadNet(name string) (int, error) {
```
iaguis: Should we expose log levels on these functions too?

Maybe setting logging is something rare enough that users should just use `Load()` explicitly.

schu: > Maybe setting logging is something rare enough that users should just use `Load()` explicitly.

I was leaning towards that. Also, in the current bcc release (0.5.0), enabling logging does not only enable verifier logging but also makes [bcc write hints to stderr](https://github.com/iovisor/bcc/blob/9634a002c694c7157fdff666522739e3eecb387b/src/cc/libbpf.c#L479-L481).

iaguis: Yeah, [I complained about that](https://github.com/iovisor/bcc/issues/1472). All good then.

PR #108, opened 2018-02-01
semaphore: update kernels to test on

iaguis:
Use the latest longterm kernels.

Also, update the rkt version used to v1.29.


schu:
Thanks @iaguis

PR #111, opened 2018-03-28
bpf_test: test `progtestrun`

schu:


schu:
Re-doing https://github.com/iovisor/gobpf/pull/110 - this time without accidentally added files.

PR #112, opened 2018-04-16
bcc.Table API: Use iterator instead of channel, provide raw byte-slices

gimpf:
Currently `bcc.Table` provides iteration using a channel, a pattern that causes high overhead, and also always converts values to and from strings, which is inconvenient. This PR changes that API to a cursor-like iterator, providing a `Next()` method for iteration, and `Key()`, `Leaf()` to access the raw byte-slices, as well as `KeyString()`, `LeafString()` to continue supporting simple printing of table contents.

Note: Given that this is already a breaking API change, we'd offer to merge parts of @nathanleclaire's PR #102, which would change the regular `Get` and `Set` methods to use byte-slices as well.

### DIFF
--- a/bcc/perf.go
+++ b/bcc/perf.go
@@ -55,15 +55,7 @@ var mu sync.Mutex
 
 // In lack of binary.HostEndian ...
 func init() {
-	var i int32 = 0x01020304
-	u := unsafe.Pointer(&i)
-	pb := (*byte)(u)
-	b := *pb
-	if b == 0x04 {
-		byteOrder = binary.LittleEndian
-	} else {
-		byteOrder = binary.BigEndian
-	}
+	byteOrder = determineHostByteOrder()
 }
 
 func registerCallback(data *callbackData) uint64 {
@@ -99,6 +91,23 @@ func callback_to_go(cbCookie unsafe.Pointer, raw unsafe.Pointer, rawSize C.int) 
 	go func() {
 		receiverChan <- C.GoBytes(raw, rawSize)
 	}()
+}
+
+// GetHostByteOrder returns the current byte-order.
+func GetHostByteOrder() binary.ByteOrder {
+	return byteOrder
+}
+
+func determineHostByteOrder() binary.ByteOrder {
+	var i int32 = 0x01020304
+	u := unsafe.Pointer(&i)
+	pb := (*byte)(u)
+	b := *pb
+	if b == 0x04 {
+		return binary.LittleEndian
+	}
+
+	return binary.BigEndian
 }
 
 // InitPerfMap initializes a perf map with a receiver channel.
--- a/bcc/table.go
+++ b/bcc/table.go
@@ -32,6 +32,7 @@ import "C"
 
 var errIterationFailed = errors.New("table.Iter: leaf for next key not found")
 
+// Table references a BPF table.  The zero value cannot be used.
 type Table struct {
 	id     C.size_t
 	module *Module
@@ -85,12 +86,6 @@ func (table *Table) LeafStrToBytes(leafStr string) ([]byte, error) {
 	return leaf, nil
 }
 
-// Entry represents a table entry.
-type Entry struct {
-	Key   string
-	Value string
-}
-
 func (table *Table) KeyStrToBytes(keyStr string) ([]byte, error) {
 	mod := table.module.p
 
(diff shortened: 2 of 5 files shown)


bcc/table.go
```diff
@@ -181,56 +185,134 @@ func (table *Table) DeleteAll() error {
...
+	fd := C.bpf_table_fd_id(mod, table.id)
+	keySize := C.bpf_table_key_size_id(mod, table.id)
+	leafSize := C.bpf_table_leaf_size_id(mod, table.id)
+	key := make([]byte, keySize)
+	leaf := make([]byte, leafSize)
+
+	nextKey := make([]byte, keySize)
+	keyP := unsafe.Pointer(&nextKey[0])
+
+	res := C.bpf_get_first_key(fd, keyP, keySize)
```
schu: I was wondering if this also could be done in `Next()`? Something like

```
if it.key == nil {
   // get first key
   return true
}
// get next key
```

Then `errNoIterationStartingPoint` wouldn't be needed and the implementation be closer to Go's `bufio.Scanner`.

bcc/table.go
```diff
@@ -181,56 +185,134 @@ func (table *Table) DeleteAll() error {
...
+func (it *TableIterator) Next() bool {
+	if it.err != nil || it.nextKey == nil {
+		return false
+	}
+
+	it.key, it.nextKey = it.nextKey, it.key
+
+	keyP := unsafe.Pointer(&it.key[0])
+	leafP := unsafe.Pointer(&it.leaf[0])
+	res := C.bpf_lookup_elem(it.fd, keyP, leafP)
```
schu: Again with `bufio.Scanner` in mind: what do you think about e.g.

```
diff --git a/bcc/table.go b/bcc/table.go
index a8c1954..b304beb 100644
--- a/bcc/table.go
+++ b/bcc/table.go
@@ -18,6 +18,7 @@ import (
        "bytes"
        "errors"
        "fmt"
+       "os"
        "unsafe"
 )
 
@@ -238,16 +239,22 @@ func (it *TableIterator) Next() bool {
 
        keyP := unsafe.Pointer(&it.key[0])
        leafP := unsafe.Pointer(&it.leaf[0])
-       res := C.bpf_lookup_elem(it.fd, keyP, leafP)
+       res, err := C.bpf_lookup_elem(it.fd, keyP, leafP)
        if res != 0 {
-               it.err = errIterationFailed
+               if !os.IsNotExist(err) {
+                       it.err = err
+               }
                return false
        }
 
        nextKeyP := unsafe.Pointer(&it.nextKey[0])
-       res = C.bpf_get_next_key(it.fd, keyP, nextKeyP)
+       res, err = C.bpf_get_next_key(it.fd, keyP, nextKeyP)
        if res != 0 {
                it.nextKey = nil
+               if !os.IsNotExist(err) {
+                       it.err = err
+                       return false
+               }
        }
 
        return true
```

I.e.

* when `Next()` returns `false` and `Err()` returns `== nil`, the end of the table is reached,
* when `Next()` returns `false` and `Err()` returns `!= nil`, the returned error occured.

schu: We could then also check that `iter.Err()` is `nil` in the tests.

gimpf:
@schu @nathanleclaire Sorry for the late update, some more urgent tasks came up. But now...

I've integrated @schu's suggestions from the CR, and also incorporated @nathanleclaire's changes to the Set/Get methods, although with some changes.

* Get now only returns the leaf, not any entry (the caller obviously already has the key value...)
* Table now exposes the sscanf and sprintf table functions for leaf and key; the might be convenient enough to use even outside of iterating all of the table

Obviously, more could be done, but I think this PR is a pareto improvement to the current master. What do you think?

P.S.: The builds fail due to some 404. AFAICS it's not caused by the changes to the code. That right?

schu:
@gimpf thanks for the update.

> The builds fail due to some 404. AFAICS it's not caused by the changes to the code. That right?

Right, the test images have been updated meanwhile and the old images are not available anymore, hence the early 404 error. Please rebase on top of master for the tests to run.

Also, before a second review, please cleanup / squash your commit history where necessary to remove fixup commits etc. 47e59031f6b8674061ed264a3ab410560ce88761 for example is a fixup for a previous commit and 07dbe032011786b7129f180a5731d48e1b6d761e shouldn't be included entirely, I think.

gimpf:
@schu As requested, history cleaned. Squashed it down to three commits.

bcc/table.go
```diff
@@ -181,56 +212,111 @@ func (table *Table) DeleteAll() error {
...
+}
+
+// Leaf returns the current leaf value of the iterator, if the most recent call to Next returned true.
+// The slice is valid only until the next call to Next.
+func (it *TableIterator) Leaf() []byte {
+	return it.leaf
+}
+
+// KeyString returns the current key value of the iterator formatted using the bcc-table's key string printer, if the most recent call to Next returned true.
+func (it *TableIterator) KeyString() (string, error) {
```
bobrik: I think users can use `KeyBytesToStr` if they want to, it's not necessary to add this to the iterator itself. Just my 2c.

bobrik:
I have a PR open to use these changes: https://github.com/cloudflare/ebpf_exporter/pull/24.

Apart from my two comments, which are not blockers, I think this is good to go.

bobrik:
@schu is there anything that needs to happen to see this merged?

gimpf:
@bobrik As it is, the code does not compile anymore, as in the meantime we upgraded to bcc 0.6.0, which has some breaking API changes. I'll fix this as soon as I find time, and also integrate your suggestions.

bcc/perf.go
```diff
@@ -101,6 +93,19 @@ func callback_to_go(cbCookie unsafe.Pointer, raw unsafe.Pointer, rawSize C.int)
 	}()
 }
 
+// GetHostByteOrder determines and returns the current byte-order.
+func GetHostByteOrder() binary.ByteOrder {
```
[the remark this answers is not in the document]
schu: > Should we ask for Go to provide host byte order out of the box? Something like `binary.HostOrder`.

I think this has been discussed in the past, for example here: https://groups.google.com/forum/#!topic/golang-nuts/3GEzwKfRRQw

gimpf:
@bobrik @schu Branch is rebased; as I agree with bobrik's assessment above, I also incorporated his suggestions. Test run on my machine. CI has issues, again.

bobrik:
Tests are passing.

bobrik:
I would really love to see this merged 👍

schu:
Thank you!

gimpf:
Thanks for accepting this PR!

nathanleclaire:
💯 🎉 🎆

Issue #113, opened 2018-04-22
Support for symbol lookup


bobrik:
In Python it's possible to do `BPF.ksym(ip)` to lookup function name by instruction pointer:

* https://github.com/iovisor/bcc/blob/581052a68e5/src/python/bcc/__init__.py#L1068-L1080

Go should be able to do the same set of lookups.

Pryz:
agree https://github.com/iovisor/gobpf/pull/105 :)

bobrik:
@Pryz did you make that helper package in the end?

schu:
>  never got any response. Let me create a PR quickly with the Ksym func in a helper package to start the discussion.

Sorry, I missed that. Feel free to ping me after a while if you don't get any response :) I'll comment on the pull request.

Pryz:
@schu no worries ! I will ping you next time ;) thanks !

schu:
Was added with #114

PR #114, opened 2018-04-23
Add ksym helper

Pryz:
This PR is based on https://github.com/iovisor/gobpf/pull/105 but adds the `Ksym` function in a helper package.

People found this kind of helpers pretty useful in the Python library. See : https://github.com/iovisor/gobpf/issues/113

cc @bobrik

### DIFF
--- a/pkg/ksym/ksym.go
+++ b/pkg/ksym/ksym.go
@@ -1,0 +1,66 @@
+package ksym
+
+import (
+	"bufio"
+	"errors"
+	"io"
+	"os"
+	"strings"
+	"sync"
+)
+
+const (
+	KALLSYMS = "/proc/kallsyms"
+)
+
+type ksymCache struct {
+	sync.RWMutex
+	ksym map[string]string
+}
+
+var cache ksymCache
+
+// Ksym translates a kernel memory address into a kernel function name
+// using `/proc/kallsyms`
+func Ksym(addr string) (string, error) {
+	if cache.ksym == nil {
+		cache.ksym = make(map[string]string)
+	}
+
+	cache.Lock()
+	defer cache.Unlock()
+
+	if _, ok := cache.ksym[addr]; !ok {
+		fd, err := os.Open(KALLSYMS)
+		if err != nil {
+			return "", err
+		}
+		defer fd.Close()
+
+		fn := ksym(addr, fd)
+		if fn == "" {
+			return "", errors.New("kernel function not found for " + addr)
+		}
+
+		cache.ksym[addr] = fn
+	}
+
+	return cache.ksym[addr], nil
+}
+
+func ksym(addr string, r io.Reader) string {
+	s := bufio.NewScanner(r)
+	for s.Scan() {
+		l := s.Text()
+		ar := strings.Split(l, " ")
+		if len(ar) != 3 {
+			continue
+		}
+
+		if ar[0] == addr {
+			return ar[2]
+		}
+	}
+
+	return ""
+}
--- a/pkg/ksym/ksym_test.go
+++ b/pkg/ksym/ksym_test.go
@@ -1,0 +1,17 @@
+package ksym
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestKsym(t *testing.T) {
+	data := "ffffffff91b2a340 T cgroup_freezing"
+
+	r := strings.NewReader(data)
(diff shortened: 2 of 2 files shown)


bobrik:
Should we make this cacheable? My understanding is that `/proc/kallsyms` should only change on module load / unload, and that's a rate event. Currently we have to re-read it for each resolve.

I can add caching in my own code for symbols I resolve, not sure if that's the right approach.

bobrik:
Hmm, something is not right and I'm not sure where exactly.

I have the following ebpf program:

```
#include <uapi/linux/ptrace.h>

struct key_t {
    u64 ip;
};
BPF_HASH(counts, struct key_t);

int do_count(struct pt_regs *ctx) {
    struct key_t key = {};
    u64 zero = 0, *val;
    u64 ip;
    key.ip = PT_REGS_IP(ctx);
    val = counts.lookup_or_init(&key, &zero);  // update counter
    (*val)++;
    return 0;
}
```

It's effectively copied from `cachestat`:

* https://github.com/iovisor/bcc/blob/master/tools/cachestat.py

In python code of `cachestat` the address is resolved to name with just `b.ksym(ip)`.

I attach `mark_page_accessed` krpobe to `do_count` and get the table with the following contents:

```
{ 0xffffffffae36b7d1 } -> 1331038
```

Here `0xffffffffae36b7d1` is supposed to be the address of `mark_page_accessed`, but I have:

```
$ cat /proc/kallsyms | fgrep mark_page_accessed
ffffffffae36b7d0 T mark_page_accessed
ffffffffaf0be0f0 r __ksymtab_mark_page_accessed
ffffffffaf0db2a1 r __kstrtab_mark_page_accessed
```

This looks like an off by one.

Pryz:
Yes we can probably cache those. I'm currently caching the specific symbols in my application.

It seems like Python is using `bcc_symcache_resolve_name` (https://github.com/iovisor/bcc/blob/8322f800226fbcf54e24a4b179e129c53d6a755a/src/cc/bcc_syms.cc#L352) and not `/proc/kallsyms` which I assumed was returning the same value.

schu:
@Pryz I'm fine with putting this in `pkg/ksym` like in the first PR :+1:  (and actually would prefer that over a new `helper` namespace). My initial comment was rather if we should put the ksym helper code into gobpf at all. 

Do you want to look into adding a cache first or do that in a later PR?

@bobrik I don't have an idea regarding the unexpected address. Do you have the test program on GitHub?

Thanks!

Pryz:
@schu sorry for the misunderstanding here :) I will move that back to `pkg/ksym` and add the cache layer in that PR.

Thanks !

bobrik:
@schu I have this: https://github.com/cloudflare/ebpf_exporter. You can try `cachestat` example.

bobrik:
Interestingly, tracepoints I added in #118 resolve ksym at the right address.

Should `PT_REGS_IP(ctx)` report address 1 byte above ksym address?

cc @4ast

schu:
> In python code of cachestat the address is resolved to name with just b.ksym(ip).

`ip` also is one above in the Python example (I'm guessing it's the addr of the kprobe breakpoint), but `b.ksym(ip)` further uses `bcc_symcache_resolve` and that [`KSyms::resolve_addr`](https://github.com/iovisor/bcc/blob/5d23500a930735aa337952db65c1155e5473aaa3/src/cc/bcc_syms.cc#L61-L83) to lookup the symbol name in a cache. I don't know if it's safe to assume the offset is always `1` (i.e. to do `PT_REGS_IP(ctx) - 1`), but the addr that you see is right.

Pryz:
Yep, sorry I'm a bit late on my schedule. Gonna try to finish that in the next few days.

schu:
Looks good to me.

As is, there's no way for the user to clear the cache or use no cache at all. I'm not sure if that is OK for all users. Not a blocker from my point of view, could be changed later if necessary. @bobrik what do you think?

Pryz:
I was thinking about adding a mutex on the map to make it concurrently safe. Will do that asap so we can finally merge this PR :)

schu:
> tests are failing but from what I see it's not due to my change. Can you take a look?

Yes, unfortunately the tests often fail currently with a qemu-kvm error (`qemu: terminating on signal 15...`). I have send a mail to SemaphoreCI support a couple of days ago and hope they can fix it.

schu:
Thanks @Pryz

PR #118, opened 2018-04-28
Add support for tracepoints, closes #117

### LINKED ISSUE #117: TRACEPOINT_PROBE doesn't work as expected
I'm looking at these docs:

* https://github.com/iovisor/bcc/blob/master/docs/reference_guide.md#3-tracepoints

Minimal reproduction:

```go
package main

import (
	"log"
	"time"

	"github.com/iovisor/gobpf/bcc"
)

var code = `
TRACEPOINT_PROBE(random, urandom_read) {
    bpf_trace_printk("%d\\n", args->got_bits);
    return 0;
};
`

func main() {
	m := bcc.NewModule(code, []string{})
	log.Printf("%#v", m)

	time.Sleep(time.Hour)
}
```

It compiles successfully:

```
2018/04/27 05:17:17 &bcc.Module{p:(unsafe.Pointer)(0xf9d380), funcs:map[string]int{}, kprobes:map[string]unsafe.Pointer{}, uprobes:map[string]unsafe.Pointer{}}
```

But I don't see it running:

```
$ sudo /usr/share/bcc/tools/bpflist
PID    COMM             TYPE     COUNT
```

If I add kprobe:

```go
package main

import (
	"log"
	"time"

	"github.com/iovisor/gobpf/bcc"
)

var code = `
TRACEPOINT_PROBE(random, urandom_read) {
    bpf_trace_printk("%d\\n", args->got_bits);
    return 0;
};

int do_count(struct pt_regs *ctx) {
    return 0;
}
`

func main() {
	m := bcc.NewModule(code, []string{})
	log.Printf("%#v", m)

	target, err := m.LoadKprobe("do_count")
	if err != nil {
		log.Fatal(err)
	}

	err = m.AttachKprobe("add_to_page_cache_lru", target)
	if err != nil {
		log.Fatal(err)
	}

	time.Sleep(time.Hour)
}
```

Then I can see the program in `bpflist`:

```
$ sudo /usr/share/bcc/tools/bpflist
PID    COMM             TYPE     COUNT
16356  tp               prog     1
```

But it doesn't make `TRACEPOINT_PROBE` work anyway.

I've tried `urandomread.py` from python examples and it works fine:

* https://github.com/iovisor/bcc/blob/a4159da8c4e/examples/tracing/urandomread.py

```
$ sudo /usr/share/bcc/examples/tracing/urandomread.py
TIME(s)            COMM             PID    GOTBITS
6014114.438438000  sshd             18027  00000001: 128
6014114.439753000  systemd          1      00000001: 128
6014114.445211000  <...>            18035  00000001: 256
```

```
$ sudo cat /sys/kernel/debug/tracing/trace_pipe
           <...>-18042 [013] .... 6014115.548416: 0x00000001: 8192
           <...>-18042 [013] .... 6014115.548430: 0x00000001: 8192
           <...>-18042 [013] .... 6014115.548436: 0x00000001: 8192
           <...>-18042 [013] .... 6014115.548442: 0x00000001: 8192
           <...>-18042 [013] .... 6014115.548448: 0x00000001: 8192
^C
```

bobrik:


bcc/module.go
```diff
@@ -231,6 +239,32 @@ func (bpf *Module) AttachKretprobe(fnName string, fd int) error {
 	return bpf.attachProbe(evName, BPF_PROBE_RETURN, fnName, fd)
 }
 
+// AttachTracepoint attaches a tracepoint fd to a function
+func (bpf *Module) AttachTracepoint(name string, fd int) error {
```
alban: Can this be documented that `name` is in the format `category:name`?

alban:
LGTM after the comment above

bcc/module.go
bobrik: Sure, added.

bobrik:
Not really sure why the build failed here.

bobrik:
Added cleanup for tracepoints on `Close()`. Test also looks happy now.

Not sure if having strings as key is a good idea. Maybe it's better to introduce something like this:

```go
type TracepointName struct {
  Category string
  Name     string
}
```

Then it can be used in parameters and as a key in map.

I started adding support for perf events and they too have category and "config", both of which are int.

bobrik:
Is there anything I could do to help get this merged?

schu:
Thanks @bobrik 

(Tests on master failed because I'm in between updating stage1 images for the tests, pull request coming.)

PR #123, opened 2018-05-12
Add support for perf events

bobrik:
Same comment applies: https://github.com/iovisor/gobpf/pull/118#issuecomment-387263354.

I have a use-case PR ready for this too: https://github.com/cloudflare/ebpf_exporter/pull/16.

### DIFF
--- a/bcc/module.go
+++ b/bcc/module.go
@@ -22,6 +22,8 @@ import (
 	"sync"
 	"syscall"
 	"unsafe"
+
+	"github.com/iovisor/gobpf/pkg/cpuonline"
 )
 
 /*
@@ -39,6 +41,7 @@ type Module struct {
 	kprobes     map[string]int
 	uprobes     map[string]int
 	tracepoints map[string]int
+	perfEvents  map[string][]int
 }
 
 type compileRequest struct {
@@ -92,6 +95,7 @@ func newModule(code string, cflags []string) *Module {
 		kprobes:     make(map[string]int),
 		uprobes:     make(map[string]int),
 		tracepoints: make(map[string]int),
+		perfEvents:  make(map[string][]int),
 	}
 }
 
@@ -136,6 +140,11 @@ func (bpf *Module) Close() {
 		C.free(unsafe.Pointer(tpCategoryCS))
 		C.free(unsafe.Pointer(tpNameCS))
 	}
+	for _, vs := range bpf.perfEvents {
+		for _, v := range vs {
+			C.bpf_close_perf_event_fd((C.int)(v))
+		}
+	}
 	for _, fd := range bpf.funcs {
 		syscall.Close(fd)
 	}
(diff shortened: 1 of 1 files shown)


bcc/module.go
```diff
@@ -137,6 +139,12 @@ func (bpf *Module) Close() {
 		C.free(unsafe.Pointer(tpCategoryCS))
 		C.free(unsafe.Pointer(tpNameCS))
 	}
+	for _, vs := range bpf.perfEvents {
+		for _, v := range vs {
+			C.perf_reader_free(v)
+		}
+		// TODO: ?
```
bobrik: Not exactly sure what needs to happen here yet.

cpg1111:
Anyone taking a look at this? I too would like this feature.  If it's a question of the cleanup, I made my own attempt that seems to work here https://github.com/cpg1111/gobpf/blob/add_perf_event_attach_detach/bcc/module.go#L144

bcc/module.go
schu: From looking at the Python implementation (https://github.com/iovisor/bcc/blob/v0.6.0/src/python/bcc/__init__.py#L810-L844), `bpf_close_perf_event_fd` to close fd's from `bpf_attach_perf_event`. So `perfEvents  map[string][]unsafe.Pointer` can be changed to `perfEvents  map[string][]int`, I think.

bobrik: Updated, please have another look

bcc/module.go
```diff
@@ -274,6 +285,37 @@ func (bpf *Module) AttachTracepoint(name string, fd int) error {
...
+
+	if cpu > 0 {
+		r, err := C.bpf_attach_perf_event(C.int(fd), C.uint(evType), C.uint(evConfig), C.ulong(samplePeriod), C.ulong(sampleFreq), C.int(pid), C.int(cpu), C.int(groupFd))
+		if r < 0 {
+			return fmt.Errorf("failed to attach BPF perf event: %v", err)
+		}
+
+		res[cpu] = int(r)
+	} else {
+		for i := 0; i < len(res); i++ {
```
schu: This doesn't work when not all CPUs are online. For example, after disabling one CPU out of four:

```
# echo 0 > /sys/devices/system/cpu/cpu1/online
# cat /sys/devices/system/cpu/online 
0,2-3
```

```
perf_event_open failed: No such device
Failed to attach perf event: failed to attach BPF perf event: no such device
```

You can use the [cpuonline](https://github.com/iovisor/gobpf/blob/master/pkg/cpuonline/cpu_range.go) package instead of `runtime.NumCPU()`.

bcc/module.go
```diff
@@ -274,6 +288,46 @@ func (bpf *Module) AttachTracepoint(name string, fd int) error {
...
+			return fmt.Errorf("failed to determine online cpus: %v", err)
+		}
+
+		for _, i := range cpus {
+			r, err := C.bpf_attach_perf_event(C.int(fd), C.uint(evType), C.uint(evConfig), C.ulong(samplePeriod), C.ulong(sampleFreq), C.int(pid), C.int(i), C.int(groupFd))
+			if r < 0 {
+				return fmt.Errorf("failed to attach BPF perf event: %v", err)
+			}
+
+			res[i] = int(r)
```
schu: In a scenario where CPUs `[0, 2, 3]` are online, this leads to a `panic: runtime error: index out of range` as r[0] - r[2] are allocated. So I think this needs to be changed to `res = append(res, int(r))`.

bobrik: Nice catch, updated.

PR #124, opened 2018-05-16
 initial no-bcc uprobe suppport

graphaelli:
This is a quick pass at uprobe support for gobpf without a dependency on libbcc.

Still to do:
* `FETCHARGS` support
* ~handle event close~

I'd appreciate any feedback on whether this effort is worthwhile and if I'm on the right path.


graphaelli:
On my machine, this fails with:

```
--- FAIL: TestModuleLoadELF (0.01s)
	bpf_test.go:472: error while loading "cgroup/skb" (invalid argument):
		0: (b7) r0 = 123
		1: (95) exit
		At program exit the register R0 has value (0x7b; 0x0) should have been 0 or 1
```

Rewind one commit (to before disabled test was reenabled) and all tests pass.

graphaelli:
Now it makes sense, that test is skipped on [semaphoreci](https://semaphoreci.com/alban/gobpf-2/branches/pull-request-124/builds/1):

```
=== RUN   TestModuleLoadELF
--- PASS: TestModuleLoadELF (0.07s)
	bpf_test.go:220: kernel doesn't support cgroup-bpf. Skipping...
	bpf_test.go:417: kernel doesn't support BPF_PROG_TEST_RUN. Skipping...
```

graphaelli:
closing in favor of #128

PR #125, opened 2018-05-17
cgroup skb must return 0 or 1

graphaelli:
as in #124:

```
--- FAIL: TestModuleLoadELF (0.01s)
	bpf_test.go:472: error while loading "cgroup/skb" (invalid argument):
		0: (b7) r0 = 123
		1: (95) exit
		At program exit the register R0 has value (0x7b; 0x0) should have been 0 or 1
```


graphaelli:
dup of # #116

PR #134, opened 2018-07-12
Add function to get a syscall function name by kernel version.

calavera:
Some kernel versions use `sys_` as the prefix for function names,
but newer kernels use `__x64_sys_`. This function can be used to load a
specific function name based in the kernel version.

For example:
```
m := bpf.NewModule(source, []string{})
defer m.Close()

fnname := m.GetSyscallFnName("execve")
```

Signed-off-by: David Calavera <david.calavera@gmail.com>

### DIFF
--- a/bcc/module.go
+++ b/bcc/module.go
@@ -471,3 +471,21 @@ func (bpf *Module) AttachXDP(devName string, fd int) error {
 func (bpf *Module) RemoveXDP(devName string) error {
 	return bpf.attachXDP(devName, -1, 0)
 }
+
+func GetSyscallFnName(name string) string {
+	return GetSyscallPrefix() + name
+}
+
+func GetSyscallPrefix() string {
+	_, err := bccResolveName("", "sys_bpf", -1)
+	if err == nil {
+		return "sys_"
+	}
+
+	_, err = bccResolveName("", "__x64_sys_bpf", -1)
+	if err == nil {
+		return "__x64_sys_"
+	}
+
+	return "sys_"
+}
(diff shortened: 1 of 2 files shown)


calavera:
I've tested this in kernel 4.13 which uses `sys_` as prefix, and in 4.17.5 which uses `__x64_sys_` and both work as expected.

calavera:
now that I look more at the package structure in this directory, should this be in the `elf` package? this is not specific to bcc.

bcc/module.go
```diff
@@ -412,6 +412,24 @@ func (bpf *Module) AttachMatchingUretprobes(name, match string, fd, pid int) err
 	return nil
 }
 
+func (bpf *Module) GetSyscallFnName(name string) string {
```
schu: Since the module is not needed, can you change `GetSyscallFnName` and `GetSyscallPrefix` like follows?

```
diff --git a/bcc/module.go b/bcc/module.go
index b312cb6..a8ccb19 100644
--- a/bcc/module.go
+++ b/bcc/module.go
@@ -412,11 +412,11 @@ func (bpf *Module) AttachMatchingUretprobes(name, match string, fd, pid int) err
        return nil
 }
 
-func (bpf *Module) GetSyscallFnName(name string) string {
-       return bpf.GetSyscallPrefix() + name
+func GetSyscallFnName(name string) string {
+       return GetSyscallPrefix() + name
 }
 
-func (bpf *Module) GetSyscallPrefix() string {
+func GetSyscallPrefix() string {
        _, err := bccResolveName("", "sys_bpf", -1)
        if err == nil {
                return "sys_"
```

calavera: good call, I made the change.

schu:
Thanks @calavera

PR #139, opened 2018-08-08
[WIP] elf: update bpf.h and add bpf_map.h

schu:
Update `bpf.h` with the latest version from Linux 4.17. Among many other
updates, it includes the `GPL-2.0 WITH Linux-syscall-note` license
exception:

https://github.com/torvalds/linux/blob/v4.17/LICENSES/exceptions/Linux-syscall-note

Use `bpf_map.h` to contain our custom `bpf_map_def` etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ sections (`SEC("...")`). Currently supported are:
 * `socket...`
 * `tracepoint...`
 
-Map definitions must correspond to `bpf_map_def` from [the elf package](https://github.com/iovisor/gobpf/blob/master/elf/include/bpf.h)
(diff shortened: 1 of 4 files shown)


dholbach:
Not sure if it matters, but :+1: from me.

After a conversation with somebody knowledgeable when it comes to licenses, what's presented in this PR was deemed a good solution. Namely it:

 - spells out which license the kernel header is under
 - adds the license text to the repo
 - includes the kernel header from a documented origin, unmodified
 - includes the added linux-syscall-note in the source
 - includes the linux-syscall-note text
 - updates from upstream kernel tree

PR #140, opened 2018-08-08
bcc: add SetP,GetP,DeleteP to Table for handling key/leaf as unsafe.Pointer

higebu:
This PR adds SetP,GetP,DeleteP to Table for handling key/leaf as unsafe.Pointer.
It is useful if we use `struct` for key or leaf of bpf table.

```
func (table *Table) GetP(key unsafe.Pointer) (unsafe.Pointer, error)
func (table *Table) SetP(key, leaf unsafe.Pointer) error
func (table *Table) DeleteP(key unsafe.Pointer) error
```

### DIFF
--- a/bcc/table.go
+++ b/bcc/table.go
@@ -156,6 +156,21 @@ func (table *Table) Get(key []byte) ([]byte, error) {
 	return leaf, nil
 }
 
+// GetP takes a key and returns the value or nil.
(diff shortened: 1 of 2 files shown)


schu:
Thanks!

higebu:
Thanks!

PR #143, opened 2018-08-27
Match code convention, snake_case to camelCase

sangyun-han:
Change snake_case to camelCase to match code convention.

Thanks.


schu:
Thanks!

PR #144, opened 2018-09-12
semaphore: update to new gobpf-ci image (bcc v0.7.0)

schu:


schu:
CI fails due to SemaphoreCI issue with qemu-kvm.

PR #147, opened 2018-09-18
Use correct prefix for syscall probes

### LINKED ISSUE #146: Update elf module with new syscall prefixes
Our kprobes based syscall framework depends on gobpf's elf module and it worked fine till kernel v4.16. Since 4.17, syscalls have been prefixed with `__x64_` which may be causing the probes to not be inserted by our framework. We need to update the elf module to attach to proper functions by dynamically selecting the syscall name based on the version. A reference is here: https://github.com/iovisor/gobpf/pull/134

You can assign this to me as I am on it

tuxology:
Kernel versions 4.17+ need `__x64__sys_` prefix for syscalls. So we patch the section names when necessary.

Fixes #146 

Tested with Kernel `v4.16` and Kernel `v4.18.5` and seems to work


alban:
I am not sure gobpf is the right place for this. kprobes are not a stable ABI in Linux, so BPF programs should be written for a specific version of the kernel. I am worried if gobpf does conversion behind the scene.

It is different in the case of gobpf/bcc where BPF programs are generated on the fly by the library (see the fix in https://github.com/iovisor/gobpf/pull/134). If the framework using gobpf/elf generates the C source from templates, those templates should generate the correct prefix based on the kernel version.

On the other hand, it would be easier for projects if gobpf provides some kind of helpers for this. A couple of ideas:

- What about a new prefix for section names such as `syscall/` and `retsyscall` in addition to `kprobe/` and `kretprobe/` where gobpf would be allowed to do conversions?
- New code should use tracepoints instead of kprobes, where there is a stronger stability guarantee. BPF attached to tracepoints exists since Linux 4.7 and the kprobes breakage is in 4.17. So if applications try tracepoints and fallback on kprobes on Linux < 4.7 when tracepoints are not available, then there would be no breakage.

iaguis:
I'd lean towards alban's 1st point or providing some kind of `GetSyscallPrefix()` function that takes into account the kernel version so library users have it easier.

alban:
I experimented with a guard: https://github.com/kinvolk/gobpf/commit/a1fff9fe49585960c614bf3c447bad19a04e63fc
But that would require the user of gobpf to list all sections that requires it. It could be a lot if the module adds a kprobe on a lot of syscalls.

I will try to do it globally instead with a new function: `m.EnableOptionCompatProbe()`

alban:
What do you think of the following patch on top of this branch? https://github.com/kinvolk/gobpf/commit/980751990116cc679350e7a9e97c3e08e7936b8a

iaguis:
I can live with that :)

tuxology:
> What do you think of the following patch on top of this branch? kinvolk@9807519

It seems reasonable to me as well. I also agree to what you say in  https://github.com/iovisor/gobpf/pull/147#issuecomment-423481893 even though elf module is quite low level and it should not be gobpf's responsibility to maintain function name incompatibility, there may be some generic cases like syscalls which are worth taking care in gobpf itself only if the gobpf consuming library demands it. Thanks a lot. this will fix some stuff for me! Also is checking for `SyS_` prefix ok or should it be `[Ss]y[sS]_` ?

calavera:
BCC provides this information for higher level APIs, I added support for it here: https://github.com/iovisor/gobpf/commit/bb06ddd55983ed1efb7451c2ea749652782827ad

If the elf part of the repo is supposed to be lower lever/independent of BCC, I believe it should implement this functionality. It can also have a compatible API, something like `elf.GetSyscallFnName`.

tuxology:
@calavera I think ideally, the `elf` module's responsibility should not be on providing helpers to give you qualified names for which you want to insert probes. Because a elf module's consumer may already have build the BPF program from the source code they have at build time independent of the use of gobpf. For example, https://github.com/ShiftLeftSecurity/traceleft/blob/master/bpf/trace_events.c#L398 But this also means the source could have been built for another kernel version (with different function names). So just for syscalls, we can (if the maintainers agree) make a concession to patch the kprobe names - but that too when the `elf` consuming project explicitly asks via https://github.com/kinvolk/gobpf/commit/980751990116cc679350e7a9e97c3e08e7936b8a#diff-59b0534a5a680fd786ae86764556258eR235

However, what i did by checking kernel version is not a good way now that I think of it - as the version string won't indicate 100% if the syscall names actually changed. My new plan is to actually go through `/proc/kallsyms` and then search for `__$ARCH_sys_$SYSCALL` pattern where `ARCH` is determined by what arch the program is running on and `$SYSCALL` is the syscall name that kprobe asked to be inserted. Maybe this is better. Is it @alban? :thinking:

calavera:
> However, what i did by checking kernel version is not a good way now that I think of it - as the version string won't indicate 100% if the syscall names actually changed. My new plan is to actually go through `/proc/kallsyms` and then search for `__$ARCH_sys_$SYSCALL` pattern where `ARCH` is determined by what arch the program is running on and `$SYSCALL` is the syscall name that kprobe asked to be inserted.

That sounds like a good plan to me.

I do like the idea of making the bpf program as portable as possible regarding syscall names :+1:

tuxology:
@alban I pushed a change:  https://github.com/iovisor/gobpf/pull/147/commits/8f5533cb07503244cfee31066aad7776ea8ea9af This basically implements a /proc/kallsyms based approach to get the exact syscall name. You can rebase your patch you did over this (https://github.com/kinvolk/gobpf/commit/980751990116cc679350e7a9e97c3e08e7936b8a) which implements the guard for compat mode, merge that in same PR and then it should make everything work :clap: I tested this on `4.18` and will test on `4.16` now.

elf/compat.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// We should search for new syscall function like "__x64__sys_open"
+	// Note the start of word boundary. Should return exactly one string
+	regexStr := `(\b__` + arch + `_[Ss]y[sS]_` + name + `\b)`
+	fnRegex := regexp.MustCompile(regexStr)
+
+	// Get kernel symbols
+	syms, err := ioutil.ReadFile("/proc/kallsyms")
+	if err != nil {
+		fmt.Println(string(syms))
```
calavera: is this print necessary? I think it can be removed

elf/compat.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"regexp"
+	"runtime"
+)
+
+func GetSyscallFnName(name string) (string, error) {
+	var arch string
+	if runtime.GOARCH == "amd64" {
```
calavera: you can implement this if chain with a switch:

```go
switch runtime.GOARCH {
  case "386": arch = "ia32"
  default: arch = "x64"
}
```

It's more compact and clear

elf/compat.go
```diff
@@ -0,0 +1,70 @@
...
+	if len(match) > 1 {
+		return "", err
+	}
+
+	// If nothing found, search for old syscall function to be sure
+	if len(match) == 0 {
+		newRegexStr := `(\b[Ss]y[sS]_` + name + `\b)`
+		fnRegex = regexp.MustCompile(newRegexStr)
+		newMatch := fnRegex.FindAllString(string(syms), -1)
+		fmt.Println(newMatch)
```
calavera: I think this print can be removed too.

elf/compat.go
```diff
@@ -0,0 +1,70 @@
...
+	syms, err := ioutil.ReadFile("/proc/kallsyms")
+	if err != nil {
+		fmt.Println(string(syms))
+		return "", err
+	}
+
+	match := fnRegex.FindAllString(string(syms), -1)
+	// This should not be possible
+	if len(match) > 1 {
+		return "", err
```
calavera: in this case `err` would be nil, should it be a useful message? maybe this check is not necessary if it's not a possible case

tuxology: Actually yes, we can remove this. It was just in case we get more than one match but thats not possible

elf/compat.go
tuxology: Thanks, will do this :+1:

tuxology:
@calavera Thanks for the review. I am also adding some unit tests now.

> Have you tried reducing the two regular expressions to only one? something like:
> regexStr := `(\b(?:__` + arch + `_)?[Ss]y[sS]_` + name + `\b)

I tried `(\b(?:__x64_)?[Ss]y[sS]_open\b)` earlier and it matches just the old syscalls so I just created another one and didn't debug this further: https://regex101.com/r/aBnh1Z/1

calavera:
> I tried `(\b(?:__x64_)?[Ss]y[sS]_open\b)` earlier and it matches just the old syscalls so I just created another one and didn't debug this further: https://regex101.com/r/aBnh1Z/1

is that because regex101.com doesn't match "all", like FindAllString does, right? You can tell because if you remove the all names, the new names actually match. Which makes me think that this implementation might be more correct, because you try to match the new names first, otherwise you'd need to check the array of matches and pick the correct one. So carry on, nothing to change. :+1:

elf/compat.go
```diff
@@ -22,33 +22,25 @@ import (
 	"runtime"
 )
 
-func GetSyscallFnName(name string) (string, error) {
+func GetSyscallFnName(name string, symFile string) (string, error) {
```
calavera: I'd leave the old signature for the public interface and use a private function for testing, that way if someone wants to use this function, they don't need to remember that the sym file is /proc/kallsyms, something like this:

```
const defaultSymFile = "/proc/kallsyms"
func GetSyscallFnName(name string) (string, error) {
  return getSyscallFnNameWithFile(name, defaultSymFile)
}

func getSyscallFnNameWithFile(name string, symFile string) (string, error) {
  // the full implementation here
```

tuxology:
@calavera Thanks for the suggestion. I made the changes

elf/compat.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"io/ioutil"
+	"regexp"
+	"runtime"
+)
+
+const defaultSymFile = "/proc/kallsyms"
+
+func GetSyscallFnName(name string) (string, error) {
```
alban: Since it is an exported function, could you document it?

Say that the `name` parameter is the bare name of the syscall such as `open` and it returns the name of the kernel function implementing it such as `sys_open`. Maybe rename the parameter to `syscallName` to directly document it in code.

elf/compat.go
```diff
@@ -0,0 +1,65 @@
...
+	if len(match) == 0 {
+		newRegexStr := `(\b[Ss]y[sS]_` + name + `\b)`
+		fnRegex = regexp.MustCompile(newRegexStr)
+		newMatch := fnRegex.FindAllString(string(syms), -1)
+
+		// If we get something like 'sys_open' or 'SyS_open', return
+		// either (they have same addr)
+		if len(newMatch) >= 1 {
+			return newMatch[0], nil
+		}
```
alban: An `else` statement is missing. If `/proc/kallsyms` is an empty file (e.g. someone did a bind mount to hide the content of `/proc/kallsyms` manually or via systemd's [InaccessiblePaths](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#ReadWritePaths=) feature), then both `match` and `newMatch` could be empty (`nil`) and it will panic on line 64.

elf/elf.go
```diff
@@ -566,6 +566,20 @@ func (b *Module) Load(parameters map[string]SectionParams) error {
...
 
+			// If Kprobe or Kretprobe for a syscall, use correct syscall prefix in section name
+			if isKprobe || isKretprobe {
+				str := strings.Split(secName, "/")
+				if (strings.HasPrefix(str[1], "SyS_")) || (strings.HasPrefix(str[1], "sys_")) {
+					name := strings.TrimPrefix(str[1], "SyS_")
+					name = strings.TrimPrefix(name, "sys_")
+					syscallFnName, err := GetSyscallFnName(name)
+					if err != nil {
+						return err
```
alban: Shouldn't we just ignore the error in that case? If `/proc/kallsyms` is not readable, we could just keep the name from the ELF file.

elf/compat_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+package elf
+
+import (
+    "path/filepath"
+	"testing"
+)
+
+func TestGetSyscallFnName(t *testing.T) {
+    fnName, err := getSyscallFnNameWithFile("open", filepath.Join("testdata", "prefix_symbols.txt"))
```
alban: I would inline the content of `prefix_symbols.txt` and `symbols.txt` in a big string constant here in `compat_test.go` so that we could easily test more without creating many text files in the git repository. Then we can test the empty file test case I mentioned before. To make that test easy to write, we could move the `ioutil.ReadFile()` in `GetSyscallFnName()` and replace `getSyscallFnNameWithFile()` by `getSyscallFnNameWithKallsyms(syscallName string, kallsymsContent string)`

tuxology: I thought about this, but then I thought that testing `getSyscallFnNameWithFile` will be closest as it also tests file opening. But I can do it this way also if not having separate files make more sense.

elf/elf.go
tuxology: Yes, makes sense

schu:
@tuxology friendly ping :)

tuxology:
@schu Hmm, I think I need to just rebase. I incorporated Alban's change requests already in last two commits

tuxology:
@alban Rebased it on `master`. Its good to go from me :+1:

iaguis:
It now passes CI. Merging. Thanks!

PR #152, opened 2018-10-10
Remove duplicate code snippet in XDP example.

jingax10:


schu:
Hey @jingax10 -- is it actually a duplicate? As far as I understand (not a networking expert :) the intention is to handle packets which have been encapsulated twice.

The [original example](https://github.com/torvalds/linux/blob/64c5e530ac2cab7e710c3bd08ef8594a016e4f9a/samples/bpf/xdp1_kern.c#L60-L77) and the [example in bcc](https://github.com/iovisor/bcc/blob/master/examples/networking/xdp/xdp_drop_count.py#L99-L116) have the same code.

jingax10:
@mauriciovasquezbernal Thanks for your quick reply. You are right. This corresponds to double tagging in 802.1Q/802.1AD. I will close PR.

/close

PR #157, opened 2018-10-26
Add support for raw tracepoints

bobrik:
Taking advantage of https://github.com/iovisor/bcc/pull/1718.

Example usage: https://github.com/cloudflare/ebpf_exporter/pull/38.

### DIFF
--- a/bcc/module.go
+++ b/bcc/module.go
@@ -36,12 +36,13 @@ import "C"
 
 // Module type
 type Module struct {
-	p           unsafe.Pointer
-	funcs       map[string]int
-	kprobes     map[string]int
-	uprobes     map[string]int
-	tracepoints map[string]int
-	perfEvents  map[string][]int
+	p              unsafe.Pointer
+	funcs          map[string]int
+	kprobes        map[string]int
+	uprobes        map[string]int
+	tracepoints    map[string]int
+	rawTracepoints map[string]int
+	perfEvents     map[string][]int
 }
 
 type compileRequest struct {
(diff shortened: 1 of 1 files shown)


bcc/module.go
```diff
@@ -140,6 +142,11 @@ func (bpf *Module) Close() {
 		C.free(unsafe.Pointer(tpCategoryCS))
 		C.free(unsafe.Pointer(tpNameCS))
 	}
+	for _, v := range bpf.rawTracepoints {
+		C.bpf_close_perf_event_fd((C.int)(v))
```
iaguis: Do we need this?

I wrote a [simple example program](https://github.com/iovisor/gobpf/blob/iaguis/example-raw-tp/examples/bcc/rawtracepoints/rawtracepoints.go) and when I Ctrl-C I see:

```
ioctl(PERF_EVENT_IOC_DISABLE) failed: Inappropriate ioctl for device
```

I see that in [`bpf_attach_raw_tracepoint`](https://github.com/iovisor/bcc/blob/v0.7.0/src/cc/libbpf.c#L1097) there's no `bpf_attach_tracing_event()`, unlike in [`bpf_attach_tracepoint`](https://github.com/iovisor/bcc/blob/v0.7.0/src/cc/libbpf.c#L1084) so I don't think we should do this.

bobrik: Good point, removed.

bobrik:
@iaguis, @schu can you have another look? Thanks!

PR #162, opened 2018-12-18
Implement GetProgramTag

bobrik:
This can be used with `perf` when `net.core.bpf_jit_kallsyms=1`
and `--kallsyms=/proc/kallsyms` are passed to `perf record`.

On newer kernels `perf top` can also accept `--kallsyms`:

* https://www.spinics.net/lists/linux-perf-users/msg07216.html

See: https://github.com/cloudflare/ebpf_exporter/issues/39

### DIFF
--- a/bcc/module.go
+++ b/bcc/module.go
@@ -161,6 +161,12 @@ func (bpf *Module) Close() {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


bobrik:
@schu can you have a look, please?

schu:
Thanks @bobrik

PR #163, opened 2019-01-02
Remove duplicate code

florianl:
Signed-off-by: Lehner Florian <dev@der-flo.net>

This PR removes duplicate code.

### DIFF
--- a/examples/bcc/xdp/xdp_drop.go
+++ b/examples/bcc/xdp/xdp_drop.go
@@ -82,15 +82,6 @@ int xdp_prog1(struct CTXTYPE *ctx) {
             return rc;
             h_proto = vhdr->h_vlan_encapsulated_proto;
     }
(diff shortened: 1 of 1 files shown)


schu:
It indeed looks like a duplicate but is intentional to handle double tagging of packets; see also https://github.com/iovisor/gobpf/pull/152

Maybe we should add a comment before the first `if (h_proto == htons(ETH_P_8021Q) ...` along the lines of

```
// While the following code appears to be duplicated accidentally,
// it's intentional to handle double tags in ethernet frames.
```

schu:
Added a comment: https://github.com/iovisor/gobpf/pull/164

florianl:
I didn't know about this. Thanks for explaining and adding the comment in https://github.com/iovisor/gobpf/pull/164

PR #166, opened 2019-01-08
elf: allow to override max_entries for BPFMap

sfluor:
Allow to override max_entries for BPFMap through BPFMapMaxEntries SectionParams attribute.

In the case where we want to store network connections in a BPF Map, what we could do would be to get the maximum number of connections that can be running on the given host and use it to set the BPF map size.

### DIFF
--- a/elf/elf.go
+++ b/elf/elf.go
@@ -369,6 +369,13 @@ func elfReadMaps(file *elf.File, params map[string]SectionParams) (map[string]*M
 
 		mapDef := (*C.bpf_map_def)(unsafe.Pointer(&data[0]))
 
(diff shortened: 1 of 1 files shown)


sfluor:
Hi ! Thanks for the review ! I fixed the typo and renamed the param to `MaxMapEntries`

sfluor:
Hi 👋 , do you have any update on that ? @schu

schu:
@sfluor hey, please squash cef010f into the previous commit (as it's a fixup for it; we usually try to avoid merging those). Otherwise looks good to me.

schu:
Thanks!

PR #167, opened 2019-01-12
elf: add implementation to loop sched progs

moolen:
**Issue:** users of the elf module api are not able to loop over sched progs. The'd have to know the program name upfront.

This PR offers an func that returns a channel that emits the sched programs included in the module. Just like the other `Iter*` funcs.

Users have the very same issue with accessing maps. Would you like to have that added aswell?


moolen:
thanks @schu! I've rebased my changes.

PR #170, opened 2019-01-23
bcc: update prog_load to new name (introduced in iovisor/bcc@e60438d)

notti:
Without this, fails to compile with current bcc master


iaguis:
Thanks for the patch!

However, we don't support bcc master but target the latest release (currently v0.8.0). We'll be happy to merge it once they tag a new release ([around the end of February, it seems](https://github.com/iovisor/bcc/pull/2149#issuecomment-457055245)).

schu:
I cherry-picked your PR into https://github.com/iovisor/gobpf/pull/175 which has other updates for bcc v0.9.0 and will merge as part of that. Thanks again!

PR #173, opened 2019-02-18
fix iovisor/bcc/issues/#1851 for Arch Linux users

JayceCao:
The fix for **Arch Linux** came from iovisor/bcc#1851.

Have tested that the script can run on `Ubuntu 18.04.2` without any modification.
Haven't tested for other distributions.

### DIFF
--- a/examples/bcc/bash_readline/bash_readline.go
+++ b/examples/bcc/bash_readline/bash_readline.go
@@ -17,6 +17,7 @@ package main
 import (
 	"bytes"
 	"encoding/binary"
+	"flag"
 	"fmt"
 	"os"
 	"os/signal"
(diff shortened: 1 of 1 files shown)


examples/bcc/bash_readline/bash_readline.go
```diff
@@ -53,7 +54,28 @@ type readlineEvent struct {
 	Str [80]byte
 }
 
+var (
+	name string
+	h bool
+)
+
+func init() {
+	flag.BoolVar(&h, "h", false, "Print this help message")
```
schu: The flag package does print help text by default for `-h` / `--help`, so that's not needed I think.

examples/bcc/bash_readline/bash_readline.go
```diff
@@ -53,7 +54,28 @@ type readlineEvent struct {
...
+}
+
 func main() {
+	flag.Parse()
+	if h {
+		flag.Usage()
+		os.Exit(1)
+	}
+	if name == "" {
+		name = "/bin/bash"
```
schu: You can set the default directly with `flag.StringVar`.

examples/bcc/bash_readline/bash_readline.go
```diff
@@ -53,7 +54,28 @@ type readlineEvent struct {
...
 }
 
+var (
+	name string
+	h bool
+)
+
+func init() {
+	flag.BoolVar(&h, "h", false, "Print this help message")
+	flag.StringVar(&name, "s", "", `Specify the location of libreadline.so library.
```
schu: Instead of `s`, please give the flag a more descriptive name, e.g. `target`.

examples/bcc/bash_readline/bash_readline.go
```diff
@@ -53,7 +54,28 @@ type readlineEvent struct {
...
+You may need specify this option when failing to run the script directly.`)
+	flag.Usage = usage
```
schu: Let's just go with the default usage message from the flag package.

PR #179, opened 2019-04-02
Add sockmap support

### LINKED ISSUE #177: Add BPF_PROG_TYPE_SK_SKB support to gobpf/elf
I want to add support for [sockmap redirect programs](https://lwn.net/Articles/731133/). I have made some progress on this locally but wanted to post my thoughts here before continuing.

gobpf/elf already allows you to create `BPF_MAP_TYPE_SOCKMAP` maps, so I was planning to add support for sections of the form `sk/skb/<sockmap_name>/{parser,verdict}` which would attach a parse or verdict program to a map called `maps/<sockmap_name>`.

I'm not sure whether to [enforce both a parser and verdict are attached](https://stackoverflow.com/questions/55446637/are-any-bpf-programs-required-for-a-sockmap) but I don't plan to do this right now and will leave that potential requirement up to the developer.

dippynark:
Adds support for [attaching parser and redirect programs to sockmaps](https://lwn.net/Articles/731133/). This fixes https://github.com/iovisor/gobpf/issues/177


elf/elf.go
```diff
@@ -555,6 +555,7 @@ func (b *Module) Load(parameters map[string]SectionParams) error {
 			isTracepoint := strings.HasPrefix(secName, "tracepoint/")
 			isSchedCls := strings.HasPrefix(secName, "sched_cls/")
 			isSchedAct := strings.HasPrefix(secName, "sched_act/")
+			isSkSkb := strings.HasPrefix(secName, "sk/skb/")
```
alban: gobpf and libbpf unfortunately diverged in the way they name sections:
https://github.com/torvalds/linux/blob/v5.0/tools/lib/bpf/libbpf.c#L2610-L2662

But when possible, could we adopt the same section names as in libbpf? That would be:
- `sk_skb/stream_parser`
- `sk_skb/stream_verdict`

tests/dummy.c
```diff
@@ -133,4 +140,16 @@ int socket__dummy(struct __sk_buff *skb)
 	return 0;
 }
 
+SEC("sk/skb/parser/dummy_sockmap")
+int parser_dummy(struct __sk_buff *skb)
```
alban: The parameter should be `struct sk_msg_md *msg`.

Ditto in `verdict_dummy`.

tests/dummy.c
```diff
@@ -81,6 +81,13 @@ struct bpf_map_def SEC("maps/dummy_array_custom") dummy_array_custom = {
 	.pinning = PIN_CUSTOM_NS,
 };
 
+struct bpf_map_def SEC("maps/dummy_sockmap") sock_map = {
+	.type = BPF_MAP_TYPE_SOCKMAP,
+	.key_size = sizeof(int),
```
alban: Does it worth giving a more realistic key struct? Such as:
```
struct sock_key {
        __u32 sip4;
        __u32 dip4;
        __u32 sport;
        __u32 dport;
};
```

More complex example on:
https://github.com/cilium/cilium/blob/506bd14/bpf/sockops/bpf_sockops.h#L21-L46

tests/dummy.c
dippynark: I think `struct __sk_buff *` is correct:
- https://github.com/torvalds/linux/blob/master/tools/testing/selftests/bpf/progs/sockmap_parse_prog.c#L16
- https://elixir.bootlin.com/linux/v4.18/source/kernel/bpf/sockmap.c#L1497

and the verifier converts `struct __sk_buff *` to the real `struct sk_buff *` at load time. Maybe I'm missing something?

tests/dummy.c
dippynark: agreed, have added your example

PR #180, opened 2019-04-11
Fix memleak in bpf_load_map() on error path

nplanel:


nplanel:
@alban maps are released during module. Close(). https://github.com/iovisor/gobpf/blame/e0d8d785d368abfb47287693c4daa1c760e2b558/elf/module.go#L724

alban:
@nplanel ok, I didn't see the Close().

But maybe line 395, in the error case:
```
if oldMap, ok := maps[name]; ok {
  return nil, fmt.Errorf("duplicate map: %q and %q", oldMap.Name, name)
```
`cm` would need to be released too.

nplanel:
@alban I fixed missed free() L154 and moved the duplicate test above

nplanel:
@alban ping

alban:
Sorry for the delay. LGTM

PR #187, opened 2019-05-16
Implement backward ring buffers

pothos:
Useful for overwritable ring buffers which hold
the last entries written. They should be stopped
before being dumped due to their nature of
overwriting their oldest entries. Backwards dumping
starts with the newest entry.

### DIFF
--- a/elf/perf.go
+++ b/elf/perf.go
@@ -30,6 +30,7 @@ import (
 #include <sys/types.h>
 #include <stdlib.h>
 #include <stdint.h>
+#include <stdio.h>
 #include <string.h>
 #include <linux/perf_event.h>
 #include <poll.h>
@@ -45,6 +46,12 @@ struct event_sample {
 struct read_state {
 	void *buf;
 	int buf_len;
+	// These two fields are for backward reading: as opposed to normal ring buffers,
+	// backward read buffers don't update the read pointer when reading.
+	// So we keep the state externally here.
+	uint64_t data_head_initialized;
+	uint64_t data_head;
+	uint64_t wrapped;
 };
 
 static int perf_event_read(int page_count, int page_size, void *_state,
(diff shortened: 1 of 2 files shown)


elf/perf.go
```diff
@@ -45,6 +46,8 @@ struct event_sample {
 struct read_state {
 	void *buf;
 	int buf_len;
+	uint64_t data_head_initialized;
+	uint64_t data_head;
```
alban: Add a comment that this two fields are for the backward read: as opposed to normal ring buffers, backward read buffers don't update the read pointer when reading. So we keep the state externally here.

elf/perf.go
```diff
@@ -139,6 +186,44 @@ func InitPerfMap(b *Module, mapName string, receiverChan chan []byte, lostChan c
...
+		// suddenly changed
+		panic(fmt.Sprintf("cannot find map %q", pm.name))
+	}
+
+	cpuCount := len(m.pmuFDs)
+	pageSize := os.Getpagesize()
+	for cpu := 0; cpu < cpuCount; cpu++ {
+		state := C.struct_read_state{}
+	ringBufferLoop:
+		for i := 0; i < 100000; i++ {
```
alban: I initially limited the amount of messages to read to 100000, to be safe. But this should be removed now. We should make sure we are not cycling infinitely in the ring buffer though.

elf/perf.go
pothos: done

elf/perf.go
pothos: Removed it as it works. It would loop endlessly if `p->type` is always `PERF_RECORD_SAMPLE` -- can that happen?

alban: It can happen if the ring buffer is full since we don't remove the events while iterating though it. We iterate with a pointer and a modulo to cycle in the ring buffer.

alban: The modulo is done in a fast way with `& (raw_size - 1)` because entries are guaranteed to be a size of a power of two

pothos: I added a check against circling in the buffer. I've implemented it a bit defensive instead of just checking if the same position is hit, it should also trigger if jumped over it.

pothos: Updated it to cover a corner case.

pothos:
Added a new commit for the SwapAndDumpBackward method from @mauriciovasquezbernal.

```
Backward perf ring buffers cannot been read and written at the same time,
they must the stopped (or paused) before reading. This causes some issues
for applications because they can lose samples while the buffer is stoppped
or paused.

This commit implements a the SwapAndDumpBackward method that creates a new
set of perf ring buffers and swaps then in the eBPF map, then eBPF programs
can continue to write samples without losing them and the user space
application can read the old perf ring buffer.

The internals of the method is composed by 4 steps:
1. create new perf ring buffer
2. save new file descriptors in perf ring map
3. dump old perf buffer
4. close old perf ring buffer
```

alban:
I initially overlooked that the 'backward' setting is separate from the 'overwritable' setting and this PR does not make the ring buffer overwritable. So currently, the ring buffers get full and cannot be written into after a while.

I am testing a fix in my branch [alban/rb-actually-overwritable](https://github.com/kinvolk/gobpf/commits/alban/rb-actually-overwritable), see commitmsg https://github.com/kinvolk/gobpf/commit/dea0cc0aba9ba5ab21ae0060ac870f2503b17fc9.

mauriciovasquezbernal:
> I initially overlooked that the 'backward' setting is separate from the 'overwritable' setting and this PR does not make the ring buffer overwritable. So currently, the ring buffers get full and cannot be written into after a while.
> 
> I am testing a fix in my branch [alban/rb-actually-overwritable](https://github.com/kinvolk/gobpf/commits/alban/rb-actually-overwritable), see commitmsg [kinvolk@dea0cc0](https://github.com/kinvolk/gobpf/commit/dea0cc0aba9ba5ab21ae0060ac870f2503b17fc9).

I did some experiments and your conclusion seems correct. I think we should add https://github.com/kinvolk/gobpf/commit/dea0cc0aba9ba5ab21ae0060ac870f2503b17fc9 to this PR.

alban:
Done in https://github.com/iovisor/gobpf/pull/245

PR #190, opened 2019-06-21
bcc: Add support for socket filter

iAklis:
Inspired by `socket(7)` and bcc tools


iAklis:
How to fix the failure?

PR #192, opened 2019-07-11
fix: Debian version regex can fail on some debian versions if the commit sha is appended to the version

sfluor:
This fixes the parsing of the `/proc/version` file for some debian versions.

This was causing `bpf_prog_load` to fail on [this check](https://elixir.bootlin.com/linux/v4.4/source/kernel/bpf/syscall.c#L630).

The fix is to add an optional group after the kernel version to match a commit hash (multiple alphanumeric characters).

### DIFF
--- a/elf/kernel_version.go
+++ b/elf/kernel_version.go
@@ -75,18 +75,23 @@ func currentVersionUbuntu() (uint32, error) {
 	return KernelVersionFromReleaseString(releaseString)
 }
 
-var debianVersionRegex = regexp.MustCompile(`.* SMP Debian (\d+\.\d+.\d+-\d+) .*`)
+var debianVersionRegex = regexp.MustCompile(`.* SMP Debian (\d+\.\d+.\d+-\d+)(?:\+[[:alnum:]]*)?.*`)
+
+func parseDebianVersion(str string) (uint32, error) {
+	match := debianVersionRegex.FindStringSubmatch(str)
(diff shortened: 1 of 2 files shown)


kevinconaway:
@alban Is this something that you would be able to take a look at ?

schu:
Hi, thanks for the PR! Two things:

Could you squash the commits (as the extra test is more an addendum to the first commit)? Also, the suffix is not a sha hash by git but a Debian version identifier, as far as I see. So I suggest a commit message along the lines of..

```
elf: fix `debianVersionRegex` for version string with `+deb...` IDs

Debian kernels can have a `+deb...` identifier appended to the kernel
version, e.g. `4.9.168-1+deb9u3`. Fix the regular expression to be able
to parse them. Add a test case as well.
```

When I run the tests they seem to fail; do you see what's wrong?

```
$ go test -v -run TestParseDebianVersion ./elf
=== RUN   TestParseDebianVersion
--- FAIL: TestParseDebianVersion (0.00s)
    kernel_version_test.go:78: expected kernel version 264616, got 264536
FAIL
FAIL	github.com/iovisor/gobpf/elf	0.002s
FAIL
```

sfluor:
Hi !

Thanks for the review and sorry for the failing test, it should be passing now (I also squashed + edited the commit message :slightly_smiling_face: )

schu:
Thanks!

PR #204, opened 2019-09-05
add missing call to unmap and disable when freeing a perf ring buffer

mauriciovasquezbernal:
Before closing the file descriptors of the perf ring buffer it is needed to
unmap and disable them.

https://github.com/iovisor/bcc/blob/d147588ebe35b7cd2b4d253a7da18bef253ea78d/src/cc/perf_reader.c#L69

### DIFF
--- a/elf/module.go
+++ b/elf/module.go
@@ -825,7 +825,23 @@ func (b *Module) closeMaps(options map[string]CloseOptions) error {
 				return fmt.Errorf("error unpinning map %q: %v", m.Name, err)
 			}
 		}
+
+		// unmap
+		for _, base := range m.bases {
+			err := syscall.Munmap(base)
+			if err != nil {
+				return fmt.Errorf("unmap error: %v", err)
+			}
+		}
+
 		for _, fd := range m.pmuFDs {
+			// disable
+			_, _, err2 := syscall.Syscall(syscall.SYS_IOCTL, uintptr(fd), C.PERF_EVENT_IOC_DISABLE, 0)
+			if err2 != 0 {
+				return fmt.Errorf("error disabling perf event: %v", err2)
+			}
+
+			// close
 			if err := syscall.Close(int(fd)); err != nil {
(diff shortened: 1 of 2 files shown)


elf/module.go
```diff
@@ -713,7 +715,26 @@ func (b *Module) closeMaps(options map[string]CloseOptions) error {
 				return fmt.Errorf("error unpinning map %q: %v", m.Name, err)
 			}
 		}
+
+		// unmap
+		mmapSize := pageSize * (m.pageCount + 1)
+
+		for _, header := range m.headers {
+			base := C.GoBytes(unsafe.Pointer(header), C.int(mmapSize))
+			err := syscall.Munmap(base)
```
alban, line 724: @mauriciovasquezbernal How does `Munmap()` work? If I read "[Turning C arrays into Go slices](https://documentation.help/Golang/cgo.html#Turning_C_arrays_into_Go_slices)" correctly, `C.GoBytes` creates a copy of the data, so `Munmap()` is called on the copy...

I tried to read the implementation of:
- [syscall.Munmap()](https://golang.org/src/syscall/syscall_unix.go#L81): it takes the address of the first element in the slice.
- [C.GoBytes()`: https://golang.org/src/cmd/cgo/out.go?h=GoBytes#L1666): it calls gobytes()
- [gobytes()](https://golang.org/src/runtime/string.go?h=gobytes#L297): it allocates a new buffer and does a mem copy.

Given the size of the memory allocation (several pages), it might be implemented with a `mmap` instead `brk/sbrk` (libc's malloc works like that for big buffers; no idea about Golang memory allocation) and this line `err := syscall.Munmap(base)` might not return an error, but it would unmap the copy and not the initial buffer...?

mauriciovasquezbernal: You're right, that wasn't worked as expected. I pushed a change that keeps a reference to the original return of the `syscall.Map` to be used in `syscall.Munmap`.

PR #206, opened 2019-09-09
Move header files in the same folder than Go files

lebauce:
When using iovisor/gobpf from a project using Go modules, the bpf.h and bpf_map.h files and not imported as they are in the `include` subdirectory.
This fix makes it possible to use gobpf in a `go.mod` file.


lebauce:
@alban Can you please have a look ? Thank you !

bboreham:
If I understand correctly, this is to resolve the same problem described at https://github.com/golang/go/issues/26366#issuecomment-405683150

:+1: from me.

alban:
LGTM  (after the conflict is resolved)

mauriciovasquezbernal:
@lebauce could you please solve the conflict?

lebauce:
@alban @bboreham Seems like I completely missed the notifications on this one. I just rebased my PR. Sorrry for the delay

lebauce:
@alban Do you think we could remove the "do-not-merge" tag ? Thanks !

mauriciovasquezbernal:
Hi @lebauce, sorry for the big delay. I believe we can't merge this PR,  as described in https://github.com/iovisor/gobpf/pull/234, some of the files we pull from the kernel expect a `linux/filename.h` structure, if we move everything into the same folder we'll have to modify all those files making it difficult to maintain.

I think we should move ahead and merge https://github.com/iovisor/gobpf/pull/217 then.

lebauce:
@mauriciovasquezbernal Thanks for your explanation. It perfectly makes sense. Can we then move forward and merge https://github.com/iovisor/gobpf/pull/217 (once the doc.go file is also added to elf/include, elf/include/linux, elf/include/uapi, elf/include/uapi/linux and elf/lib) ?

PR #207, opened 2019-09-23
Fix bpf_attr initialization

dmitrivereshchagin:
Using plain initialization list for a union with nested anonymous
structs works incorrectly with Clang.  `syscall()` in `bpf_update_element()`
may fail if `bpf_attr` is improperly initialized.  Initialization with
`memset()` is more portable.


alban:
Thanks!

PR #208, opened 2019-10-06
elf: include gobpf's bpf.h instead of host's

alban:
To enable build on old build servers with an outdated `/usr/include/linux/bpf.h`, use gobpf's copy instead.

### DIFF
--- a/elf/table.go
+++ b/elf/table.go
@@ -25,7 +25,7 @@ import (
 )
 
 /*
-#include <linux/bpf.h>
+#include "include/bpf.h"
 #include <linux/unistd.h>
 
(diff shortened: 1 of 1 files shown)


leodido:
Hey Alban would be cool to have this configurable. WDYT?

alban:
> Hey @alban would be cool to have this configurable. WDYT?

What's the use case for using the one from the host? If the one in gobpf is outdated, I would prefer that someone updates it rather than picking the host one.

Note that this PR is about the `elf` part of the library, so the BPF program is not compiled on the fly when running the application.

Issue #156, opened 2018-10-24
gobpf: Please make perf ring buffer size configurable


somersf:
As per https://github.com/iovisor/bcc/pull/997 but for the Go API.

In bcc/perf.go, a page count parameter can't be supplied; it uses an fixed constant size.
```
const BPF_PERF_READER_PAGE_CNT = 8
...

func bpfOpenPerfBuffer(cpu uint, callbackDataIndex uint64) (unsafe.Pointer, error) {
	cpuC := C.int(cpu)
	reader, err := C.bpf_open_perf_buffer(
		(C.perf_reader_raw_cb)(unsafe.Pointer(C.callback_to_go)),
		nil,
		unsafe.Pointer(uintptr(callbackDataIndex)),
		-1, cpuC, BPF_PERF_READER_PAGE_CNT)
	if reader == nil {
		return nil, fmt.Errorf("failed to open perf buffer: %v", err)
	}
	return reader, nil
}
```

Thanks.

PR #218, opened 2019-11-18
Configurable Page Count when creating PerfMaps

jessysaurusrex:
This PR introduces a new exported function InitPerfMapWithPageCnt which allows programs that use this library to configure the size of the ring buffer.  The approach of using a new method is to not break API stability with older clients.

Addresses #156
References: https://github.com/iovisor/bcc/pull/997

### DIFF
--- a/bcc/perf.go
+++ b/bcc/perf.go
@@ -46,6 +46,7 @@ type callbackData struct {
 	receiverChan chan []byte
 }
 
+// BPF_PERF_READER_PAGE_CNT is the default page_cnt used per cpu ring buffer
 const BPF_PERF_READER_PAGE_CNT = 8
 
 var byteOrder binary.ByteOrder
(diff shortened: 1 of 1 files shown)


bcc/perf.go
```diff
@@ -186,13 +192,17 @@ func (pm *PerfMap) poll(timeout int) {
 	}
 }
 
-func bpfOpenPerfBuffer(cpu uint, callbackDataIndex uint64) (unsafe.Pointer, error) {
+func bpfOpenPerfBuffer(cpu uint, callbackDataIndex uint64, pageCnt int) (unsafe.Pointer, error) {
+	if (pageCnt & (pageCnt - 1)) != 0 {
+		return nil, fmt.Errorf("pageCnt must be divisible by 2: %d", pageCnt)
```
alban: The error message should be "pageCnt must be a power of 2"

itaysk:
Hi, is there anything preventing this PR from merge? happy to offer my help if the original author is not available (although it's lgtm)

alban:
I think the error message should be fixed as I mentioned above, but this can be done in a following up PR if the original author is not available. I'll merge it.

Issue #213, opened 2019-11-07
elf: export nowNanoseconds()?


mythi:
This would be a useful function to be exported. 

Example use-case:
BPF code sets a "data updated" timestamp using `bpf_ktime_get_ns()` and at the time of reading the data in Go code, I could say "how long ago the data was updated" using `nowNanoseconds() - bpf_timestamp`.

mythi:
any thoughts?

iaguis:
Exporting it seems reasonable. Do you think you can make a PR? I'll be happy to review it.

mythi:
@iaguis #220 submitted

PR #220, opened 2019-12-02
elf: export NowNanoseconds

mythi:
NowNanoseconds() is a useful helper func in apps that track, e.g.,
time delta to a time stamp set (using bpf_ktime_get_ns()) by eBPF.

Fixes: #213


mythi:
@iaguis PTAL?

PR #221, opened 2019-12-03
elf - remove extraneous whitespace from debianVersionRegex

ti-mo:
Here are two examples of Stretch and Buster version strings, respectively: https://regexr.com/4q232.
The regex does not match a segment like `SMP Debian 4.19.37-5+deb10u2`.

Is this worth refactoring so it can be unit tested?


schu:
#192 was merged (with a test) and should fix the issue. Thanks!

PR #225, opened 2019-12-18
bugfix: Parsing of kernel version in debian failed

yurado:
Debian kernel version looks like this: 4.9.189-3+deb9u1, the current regex didn't get a match on strings like this.


schu:
Thanks. A fix for this is in the works already with #192

schu:
#192 was merged, should be fixed.

PR #228, opened 2019-12-30
Xdp flags

walrobin2:
This change adds ability to specify XDP flags with attach.  It also adds option to specify the log size used during load of bpf programs, as the default size is not sufficient for larger bpf programs.


walrobin2:
Seems the semaphore check is failing on all pull requests.

juise:
Very useful fix, +1

PR #188, opened 2019-05-29
Support for XDP programs in ELF object files

rabbitstack:
This PR brings the support for XDP programs in ELF object files through `SEC(xdp/...)` annotations.
It allows for definition of multiple XDP programs in the same C source file. Once ELF module is instantiated/loaded, you can call `AttachXDP` passing the network device and the name of the XDP program you would like to attach on the interface. The XDP programs can be unloaded from the network device via `RemoveXDP` method.

PR #233, opened 2020-02-25
tests: update objects files

mauriciovasquezbernal:
The objects are behind the .c source file, tests are failing locally in my
machine.

See https://github.com/iovisor/gobpf/pull/188.


alban:
Is it possible to remove the object files from git, and let the tester or the CI run `tests/build` instead?

mauriciovasquezbernal:
@alban I would be possible in the CI, given that we are facing problems with it I'd suggest to merge this PR as it is and open an issue to keep track of your request.

PR #222, opened 2019-12-13
elf: fix cgo include paths

mythi:
On some systems the cgo processing fails with missing headers:

lib/nlattr.c:10:10: fatal error: include/nlattr.h: No such file or directory
10 | #include "include/nlattr.h"
   |          ^~~~~~~~~~~~~~~~~~
compilation terminated.

This commit adds the header path to cgo CFLAGS that then gets passed
to gcc.

Signed-off-by: Mikko Ylinen <mikko.ylinen@intel.com>

PR #217, opened 2019-11-14
Add doc.go so go mod keeps elf/include folder

tamalsaha:
xref: https://github.com/golang/go/issues/26366#issuecomment-405683150

Signed-off-by: Tamal Saha <tamal@appscode.com>

PR #215, opened 2019-11-10
semaphore: update to new gobpf-ci image (bcc v0.11.0)

schu:
Looking into a fix for our CI setup separately.

Issue #150, opened 2018-09-28
CI builds are broken due to Semaphore issue


schu:
```
KVM: entry failed, hardware error 0x0
EAX=00000000 EBX=00000000 ECX=00000000 EDX=00000663
ESI=00000000 EDI=00000000 EBP=00000000 ESP=00000000
EIP=0000fff0 EFL=00000002 [-------] CPL=0 II=0 A20=1 SMM=0 HLT=0
ES =0000 00000000 0000ffff 00009300
CS =f000 ffff0000 0000ffff 00009b00
SS =0000 00000000 0000ffff 00009300
DS =0000 00000000 0000ffff 00009300
FS =0000 00000000 0000ffff 00009300
GS =0000 00000000 0000ffff 00009300
LDT=0000 00000000 0000ffff 00008200
TR =0000 00000000 0000ffff 00008b00
GDT=     00000000 0000ffff
IDT=     00000000 0000ffff
CR0=60000010 CR2=00000000 CR3=00000000 CR4=00000000
DR0=0000000000000000 DR1=0000000000000000 DR2=0000000000000000 DR3=0000000000000000 
DR6=00000000ffff0ff0 DR7=0000000000000400
EFER=0000000000000000
Code=00 66 89 d8 66 e8 0c ad ff ff 66 83 c4 0c 66 5b 66 5e 66 c3 <ea> 5b e0 00 f0 30 36 2f 32 33 2f 39 39 00 fc 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
qemu: terminating on signal 15 from pid 2634 (timeout)
```

xref https://github.com/newtools/ebpf/issues/67

schu:
I was told by Semaphore support that they plan to deploy a fix in the coming days.

schu:
Seems to be fixed.

schu:
Seems to be fixed.

schu:
Unfortunately, the bug appears to be back. See recent builds.

schu:
I plan to look for and test an alternative.

jnach:
@schu Verdict?

schu:
We resorted to GitHub actions for the time being. (https://github.com/iovisor/gobpf/commit/493dbd89a11293255489e5eb11f27e1468de47b1)

PR #234, opened 2020-02-27
Fix compilation in older kernels

mauriciovasquezbernal:
This PR tries to fix the compilation and tests for different kernel versions:

- Take netlink.h and if_link.h from upstream
- Conditionally compile XDP tests according to kernel version
- Update CI image

It supersedes https://github.com/iovisor/gobpf/pull/215 & https://github.com/iovisor/gobpf/pull/222.

It can be tested locally by running `sudo ./semaphore.sh`. CI is not still back because of https://github.com/iovisor/gobpf/issues/150.

There is still an issue that is not solved and it's the support for go mod, the .h and .c files are not downloaded because they are not in the same folder as the go files. Currently we have two different solutions for that https://github.com/iovisor/gobpf/pull/217 & https://github.com/iovisor/gobpf/pull/206. 
https://github.com/iovisor/gobpf/pull/206 looks like the best according to https://go-review.googlesource.com/c/go/+/125297/5/src/cmd/cgo/doc.go, but I don't think after this PR we are  able to move all headers to a single folder because many files on the system expect a `linux/...` folder structure, said that I think we should go with https://github.com/iovisor/gobpf/pull/217.

Since you folks have been involved in similar PRs I'd like to get your feedback on this. @lebauce  @bboreham @tamalsaha @schu @mythi @alban


mauriciovasquezbernal:
> I don't understand this part - can you clarify?
> 

Files pulled from upstream use `<linux/bpf.h>`, my message meant that we should keep using it and avoid changing it to `"linux/bpf.h"`. I removed that from the commit message because it was out of context and difficult to understand.

Issue #202, opened 2019-09-04
go tests failing because change of defination of function in latest bcc master branch


devidasjadhav:
bcc commit "Support for hardware offload" d147588ebe35b7cd2b4d253a7da18bef253ea78d

added extra argument "const char *dev_name" which is failing this test.
 
```
# go test -tags integration -v ./...
# github.com/iovisor/gobpf/bcc
  ../github.com/iovisor/gobpf/bcc/module.go:98:40: not enough arguments in call to _Cfunc_bpf_module_create_c_from_string
	have (*_Ctype_char, number, **_Ctype_char, _Ctype_int, _Ctype__Bool)
	want (*_Ctype_char, _Ctype_uint, **_Ctype_char, _Ctype_int, _Ctype__Bool, *_Ctype_char)
FAIL	gobpf [build failed]
# gobpf/bcc
bcc/module.go:98:40: not enough arguments in call to _Cfunc_bpf_module_create_c_from_string
	have (*_Ctype_char, number, **_Ctype_char, _Ctype_int, _Ctype__Bool)
	want (*_Ctype_char, _Ctype_uint, **_Ctype_char, _Ctype_int, _Ctype__Bool, *_Ctype_char)
```

vincentmli:
I came across the same problem, do you have patch to fix it? still new to golang, not sure how to fix it myself
<pre>
go test -tags integration -v ./... github.com/iovisor/gobpf/bcc
/usr/local/go/src/github.com/iovisor/gobpf/bcc/module.go:98:8: error: not enough arguments
  c := C.bpf_module_create_c_from_string(cs, 2, (**C.char)(&cflagsC[0]), C.int(len(cflagsC)), (C.bool)(true))
        ^
FAIL	_/usr/src/gobpf [build failed]
# _/usr/src/gobpf/bcc
bcc/module.go:98:8: error: not enough arguments
  c := C.bpf_module_create_c_from_string(cs, 2, (**C.char)(&cflagsC[0]), C.int(len(cflagsC)), (C.bool)(true))
        ^
</pre>

how to add the "const char *dev_name" in C.bpf_module_create_c_from_string?

devidasjadhav:
Add nil as last argument.
    C.bpf_module_create_c_from_string(cs, 2, (**C.char)(&cflagsC[0]), C.int(len(cflagsC)), (C.bool)(true),nil)

PR #210, opened 2019-10-17
bcc: update bpf_module_create_c_from_string for bcc 0.11.0 (fixes #202)

alexforster:
cc @bobrik

PR #237, opened 2020-04-08
fix bpf_module_create_c_from_string with remove last arguments

xing0821:
fix bpf_module_create_c_from_string with libbcc version 0.10.0-34.git.6836168.


mauriciovasquezbernal:
Hi @xing0821,

This PR will cause a regression as this was updated in https://github.com/iovisor/gobpf/pull/210 to fix https://github.com/iovisor/gobpf/issues/202.

In order to use gobpf with BCC you'll have to update to a more recent version, or switch to an old version of gobpf.

Thanks!

PR #242, opened 2020-05-19
cache syscall prefix

itaysk:
The syscall prefix cannot change on same system, so it's wasteful to re-calculate it.

### DIFF
--- a/bcc/module.go
+++ b/bcc/module.go
@@ -523,16 +523,16 @@ func GetSyscallFnName(name string) string {
 	return GetSyscallPrefix() + name
 }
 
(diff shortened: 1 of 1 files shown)


itaysk:
In [Tracee](https://github.com/aquasecurity/tracee) we were looking up dozens of syscall names on init and were reaching ~2GB of ram because of this + another bug I fixed here: https://github.com/iovisor/gobpf/pull/241. After both fixes we are down to ~100MB, so yes very noticeable RAM consumption, and slight time to load improvement (~5 seconds)

PR #244, opened 2020-05-28
Add doc.go so go mod keeps this folder

lebauce:
Co-authored-by: Tamal Saha <tamal@appscode.com>

### DIFF
--- a/elf/elf.go
+++ b/elf/elf.go
@@ -31,6 +31,8 @@ import (
 	"syscall"
 	"unsafe"
 
+	_ "github.com/iovisor/gobpf/elf/include"
+	_ "github.com/iovisor/gobpf/elf/include/uapi/linux"
 	"github.com/iovisor/gobpf/pkg/bpffs"
 	"github.com/iovisor/gobpf/pkg/cpuonline"
 )
--- a/elf/include/doc.go
+++ b/elf/include/doc.go
@@ -1,0 +1,1 @@
+package include
--- a/elf/include/libbpf.h
+++ b/elf/include/libbpf.h
@@ -14,6 +14,8 @@
 #include <stdint.h>
 #include <linux/netlink.h>
 
+int bpf_set_link_xdp_fd(int ifindex, int fd, __u32 flags);
+
 enum libbpf_errno {
 	__LIBBPF_ERRNO__START = 4000,
 
--- a/elf/include/uapi/linux/doc.go
+++ b/elf/include/uapi/linux/doc.go
@@ -1,0 +1,1 @@
+package linux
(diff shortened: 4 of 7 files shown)


elf/lib/netlink.c
```diff
@@ -8,10 +8,12 @@
...
 #include <errno.h>
 #include <time.h>
+#include <stdbool.h>
+#include <stdio.h>
 
 #include <linux/bpf.h>
-#include "libbpf.h"
-#include "nlattr.h"
+#include "../include/libbpf.h"
+#include "../include/nlattr.h"
```
mauriciovasquezbernal, line 16: Are these changes needed? The following line should make it work without the include. Are you facing any issue without it?

https://github.com/iovisor/gobpf/blob/f718482cf42f1c9b336c6ab6775b684d74184d1c/elf/module.go#L36

lebauce: The problem happens when Go tries do compile the `elf/lib` package:
```# go build -v ./...
github.com/iovisor/gobpf/elf/lib
# github.com/iovisor/gobpf/elf/lib
netlink.c:15:10: fatal error: libbpf.h: No such file or directory
   15 | #include "libbpf.h"
      |          ^~~~~~~~~~
compilation terminated.
```

I tried to add `#cgo CFLAGS: -I${SRCDIR}/include/uapi -I${SRCDIR}/include` to `elf/lib/doc.go` but it doesn't seem to change anything. It seems Golang automatically tries to compile the netlink.c and nlattr.c files. If I change the extension from `.c` to `.src`, it runs fine.

mauriciovasquezbernal: I did some experiments locally and I think I found a solution for this problem.
1. Move the `*.c` files out from `lib` and put them at the same level of the go files.
2. Remove these two lines https://github.com/iovisor/gobpf/blob/49b58e11a4b5d6c2f091c385d416f8d6f1d0cca8/elf/module.go#L56-L57
3. Add `doc.go` to the include folder as you did in this PR
4. Don't update any include path
5. Take this declaration to elf/include/libbpf.h https://github.com/torvalds/linux/blob/75caf310d16cc5e2f851c048cd597f5437013368/tools/lib/bpf/libbpf.h#L465 (or pull the whole fine, I didn't try it)

I pushed those changes to https://github.com/iovisor/gobpf/commit/514af636d2307dabe6a042817346b919d590cae6, feel free to take them and update the PR if it works for you.

Thanks!

lebauce: Thanks for your reply.
I just updated the PR with your changes. I moved the declaration line `int bpf_set_link_xdp_fd(int ifindex, int fd, __u32 flags);` to elf/module.go because moving it to elf/include/libbpf.h did not avoid the implicit declaration warning.

lebauce:
ping @schu

PR #247, opened 2020-05-29
Add CONTRIBUTING.md

schu:


schu:
I would like to avoid a long list/text but think this are a few things that we aim for and often are not clear.

PR #270, opened 2021-01-22
add bcc socket filter support and example

zhangbo1882:


examples/bcc/protocol_count/net_protocol.go
```diff
@@ -0,0 +1,105 @@
...
+	"io/ioutil"
+	"syscall"
+	"time"
+	"unsafe"
+
+	"github.com/iovisor/gobpf/bcc"
+)
+
+const (
+	PROTOCOL_FUNC  = "protocol_count"
```
halfrost, line 16: How can I get this name list? I think 'protocol_count' is just like a filter's name, right？

zhangbo1882: It is a function which is defined in net_protocol.c.

PR #274, opened 2021-02-02
Use exact width integer argument to lostCallback. 

### LINKED ISSUE #272: build failure on Fedora armv7hl
Hi, the [oci-seccomp-bpf-hook](https://github.com/containers/oci-seccomp-bpf-hook) depends on this repo and I noticed a build failure on Fedora armv7hl. The error message can be seen here: https://github.com/containers/oci-seccomp-bpf-hook/issues/72

@dbenoit17 sent us this patch which resolves the build issue.
```
diff --git a/vendor/github.com/iovisor/gobpf/bcc/perf.go b/vendor/github.com/iovisor/gobpf/bcc/perf.go
index a9881a4..b0414d3 100644
--- a/vendor/github.com/iovisor/gobpf/bcc/perf.go
+++ b/vendor/github.com/iovisor/gobpf/bcc/perf.go
@@ -95,7 +95,7 @@ func rawCallback(cbCookie unsafe.Pointer, raw unsafe.Pointer, rawSize C.int) {
 } //export lostCallback
-func lostCallback(cbCookie unsafe.Pointer, lost C.ulong) {
+func lostCallback(cbCookie unsafe.Pointer, lost C.uint64_t) {
 	callbackData := lookupCallback(uint64(uintptr(cbCookie)))
 	if callbackData.lostChan != nil {
 		callbackData.lostChan <- uint64(lost)
```

PTAL and I guess @dbenoit17 can make a PR when he gets the chance (I'm just the middleman).

HTH.

dbenoit17:
Fixes #272 

`lostCallback` is exported via CGO and re-imported elsewhere as a C function.
The second argument is declared as uint64_t when re-imported, but declared
as unsigned long in the original Go function definition.  These two types
are not the same size on all architectures, and cause build failures on
systems where the types mismatch.

This commit changes the original function definition of lostCallback to
use the type C.uint64_t instead of C.ulong for the second argument, to
maintain compatibility across architectures.

### DIFF
--- a/bcc/perf.go
+++ b/bcc/perf.go
@@ -95,7 +95,7 @@ func rawCallback(cbCookie unsafe.Pointer, raw unsafe.Pointer, rawSize C.int) {
 }
 
 //export lostCallback
-func lostCallback(cbCookie unsafe.Pointer, lost C.ulong) {
+func lostCallback(cbCookie unsafe.Pointer, lost C.uint64_t) {
 	callbackData := lookupCallback(uint64(uintptr(cbCookie)))
 	if callbackData.lostChan != nil {
 		callbackData.lostChan <- uint64(lost)


schu:
Thanks!

PR #281, opened 2021-02-25
Fix memory corruption when reading PERCPU_XXX maps

L3n41c:
When accessing an eBPF map element, `gobpf` currently allocates a slice to store one value of that map.
It then calls `bpf_lookup_elem()` to get the element of the map.

But in the case of `PERCPU_XXX` maps, `bpf_lookup_elem()` writes an *array* of values with as many elements as the number of possible CPUs.
This led to an out-of-bounds write.

### DIFF
--- a/bcc/table.go
+++ b/bcc/table.go
@@ -20,11 +20,14 @@ import (
 	"fmt"
 	"os"
 	"unsafe"
+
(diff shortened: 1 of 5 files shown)


schu (approved):

schu:
Thanks!

PR #283, opened 2021-03-01
remove undefined:utsnameStr error on linux,386 arch

themighty1:


themighty1:
Without this, it is impossible to compile on 32-bit linux.

schu:
Thanks!

PR #310, opened 2022-01-25
add OpenPerfEvent API so that tables can be configured to catch perf event counters in bpf programs

rootfs:
`open_perf_event` is well used in bcc Python libray, this PR is to get it here for golang


rootfs:
@schu your input is highly appreciated, thanks

PR #311, opened 2022-07-05
Bcc func load attachtype for bcc > 0.24.0

chantra:
[bcc] Support for new `bcc_func_load` signature.

In https://github.com/iovisor/bcc/commit/ffff0edc00ad249cffbf44d855b15020cc968536
a new parameter was added to control the attach type. Subsequently, in https://github.com/iovisor/bcc/commit/815d1b84828c02ce10e1ea5163aede6b5786ba38,
`-1` was used to signal the default behaviour.

This change adds this new parameter to the call to `C.bcc_func_load`
when LIBBCC_VERSION is greater than 0.24.0, the last cut version which
did not introduce this change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
           # Use release 9 of llvm etc. - later versions have an unfixed
           # bug on Ubuntu:
           # https://github.com/iovisor/bcc/issues/2915
-          sudo apt-get -y install bison build-essential cmake flex git libelf-dev libfl-dev libedit-dev libllvm9 llvm-9-dev libclang-9-dev python zlib1g-dev
+          sudo apt-get -y install bison build-essential cmake flex git libelf-dev libfl-dev libedit-dev libllvm11 llvm-11-dev libclang-cpp11-dev libclang-common-11-dev libclang1-11 libclang-11-dev python zlib1g-dev
           pushd /tmp
           git clone --depth 1 --branch v0.20.0 https://github.com/iovisor/bcc.git
           mkdir -p bcc/build; cd bcc/build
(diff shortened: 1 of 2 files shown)


chantra:
This patch will be needed once BCC cuts a new release and gobpf wants to support it.

Currently, there is no good way to use the right function signature at compile time. All that BCC provide is `LIBBCC_VERSION`, which is a string and cannot be compared easily.

What is the stand of gobpf in term of bcc library support? Do we want to have gobpf backward compatible? If this is the case, having BCC expose major/minor/patch would probably help as we could add a wrapper for go to call one variant or the other depending on version.

chantra:
https://github.com/iovisor/bcc/pull/4091 made it upstream, so this should be good as it is.

chantra:
@schu what do you think of this PR? Anything you would like me to address?

muhammednagy:
Would love to see this get merged, Thank you @chantra

chantra:
bcc v0.25.0 was released. I have updated this PR to reflect it.

@schu this should be good to go.

daurnimator:
Will this solve https://github.com/iovisor/gobpf/issues/314 ?

martinetd:
> Will this solve https://github.com/iovisor/gobpf/issues/314 ?

yes, it should.
Also waiting for a release here before we upgrade bcc in nixpkgs :)

daurnimator:
@schu could you review this?

martinetd:
hm, @schu does not appear to be a member of https://github.com/orgs/iovisor/people anymore; and there have not been any other committer recently. (Sorry if he has access to the repository in any other way, github does not seem to provide a way of checking if he still has commit access)

please forgive the mass notification -- @iovisor (iirc this notifies everyone in the group) -- is there anyone willing to shepherd gobpf? What would the way forward be here if not?
(We can't upgrade bcc in nixpkgs because we need a gobpf upgrade with this first)

martinetd:
Thank you @4ast!

PR #324, opened 2023-02-27
elf: Add utsname support for riscv64

Ast-x64:
According to https://go.dev/src/syscall/ztypes_linux_riscv64.go#L573 we should use uint8 on riscv64.


petterreinholdtsen:
This patch is required to get opensnitch building for riscv64 on Debian.  It would be great if it was provided here by upstream.

cavokz:
Applied this to the Debian package: https://tracker.debian.org/news/1641951/accepted-golang-github-iovisor-gobpf-020-10-source-into-unstable.

